PR #5, opened 2019-09-23
Feed gramtools + Fix #4

### LINKED ISSUE #4: Tests fail if compiler reorders alleles
Either we need to sort the alleles ( I thought we did, so I need to check), or we need tests to be agnostic to allele order so that they do not fail.

bricoletc:
Hello rachel!

I'm adding support for nesting in gramtools and am looking to use your make_prg module as a dependency.

I'm proposing a few things here, please give me feedback on which ones you're not OK with and we can work together to (or I can) change again

1) Small changes to the command line interface: i've made it such that by default if you run the script again on the same MSA, it will overwrite the existing PRG. You can pass `no-overwrite` flag to avoid this. This makes it easier to debug interactively when you run several times
2) Did a refactoring of some functions into a `utils.py` module for ease of reading & modifying
3) The `write_prg` function writes the prg String using whitespace delimiters. I've *added* serialisation into a binary integer vector, which I read in gramtools. 
4) I believe i've fixed #4 by sorting the variants inside `get_interval_seqs` function; i changed the tests accordingly (sort the variants in a site)

### DIFF
--- a/make_prg_from_msa.py
+++ b/make_prg_from_msa.py
@@ -518,7 +518,7 @@ def main():
         if os.path.isdir(args.output_prefix):
             prefix = os.path.join(args.output_prefix, os.path.basename(args.MSA))
         else:
-            prefix = args.prefix
+            prefix = args.output_prefix
     prefix += ".max_nest%d.min_match%d" % (args.max_nesting, args.min_match_length)
 
     if args.verbose:
--- a/make_prg_from_msa.py
+++ b/make_prg_from_msa.py
@@ -36,7 +36,7 @@ def get_interval_seqs(interval_alignment):
         logging.warning("Sequences were", " ".join(list(remove_duplicates([str(record.seq).replace('-', '').upper() for record in interval_alignment]))))
         logging.warning("Using these sequences anyway, and should be ignored downstream")
         seqs = list(remove_duplicates([str(record.seq).replace('-', '').upper() for record in interval_alignment]))
-    return list(set(seqs))
+    return sorted(list(set(seqs)))
 
 class AlignedSeq(object):
     """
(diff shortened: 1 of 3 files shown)


make_prg_from_msa.py
```diff
@@ -598,48 +500,57 @@ def main():
     parser.add_argument("-f", "--alignment_format", dest='alignment_format', action='store', default="fasta",
                         help='alignment_Format of MSA, must be a biopython AlignIO input alignment_format. See '
                              'http://biopython.org/wiki/AlignIO. Default: fasta')
-    parser.add_argument("--max_nesting", dest='max_nesting', action='store', type=int, default=5,
-                        help='Maximum number of levels to use for nesting. Default: 5')
+    parser.add_argument("--max_nesting", dest='max_nesting', action='store', type=int, default=10,
```
mbhall88: In the parameter sweep I did for my simulations, we found 5 performed best.

bricoletc: Done

bricoletc:
@rmcolq is this OK to merge?

rmcolq:
Just reading through

test_make_prg.py
```diff
@@ -30,17 +30,17 @@ def test_answers():
...
     aseq = AlignedSeq("test/contains_n_and_RYKMSW.fa")
-    assert aseq.prg == "AAACG 5 T 6 C 5 GGTT"
+    assert aseq.prg == "AAACG 5 C 6 T 5 GGTT"
 
     aseq = AlignedSeq("test/contains_n_and_RYKMSW_no_variants.fa")
     assert aseq.prg == "AAACGTGGTT"
 
-    with pytest.raises(Exception):
-        aseq = AlignedSeq("test/fails.fa")
+    #with pytest.raises(Exception):
```
rmcolq: Why does this test no longer fail?

bricoletc: Due to 0552d44  we now accept full colum of N's

PR #8, opened 2019-11-08
Refactoring and improving test coverage

mbhall88:
This PR covers a couple of things sorry

- Reformat codebase with `black`
- Add a tested function for removing gaps from a sequence as this is a very common operation
- Remove redundant function `contains_only`
- Update CLI help messages to programmatically show default values
- Rename test files so they can be auto-discovered 
- Refactor `IntegerEncoder` and add tests to cover the whole class. (fixing some bugs in the process)


martinghunt:
@mbhall88  I'm curious - how do you run the tests?  I've was using `python3 setup.py test`, which doesn't care if the files are called `test_*.py` or `*_test.py`. Am I ignoring some convention by using `*_test.py`?

mbhall88:
> @mbhall88 I'm curious - how do you run the tests? I was using `python3 setup.py test`, which doesn't care if the files are called `test_*.py` or `*_test.py`. Am I ignoring some convention by using `*_test.py`?

I was using PyCharm, which effectively runs `python -m unittest`. Just found that [by default](https://docs.python.org/3.5/library/unittest.html#cmdoption-unittest-discover-p), `unittest` looks for `test_*.py`. I guess I could change the options to also search for `*_test.py`.

In general, I tend to use `pytest` for my test suite. By default, [`pytest` will look for both conventions](https://docs.pytest.org/en/latest/goodpractices.html#test-discovery).

PR #10, opened 2019-11-11
Add fasta header to PRG file

### LINKED ISSUE #9: Make output fasta formatted
The current `.prg` output file is not in a "useable" file format. I find myself having to add a fasta header after running `make_prg` (which can be a pain in a pipeline).

mbhall88:
1. Write fasta-style header to `.prg` file [closes #9 ]. The way the header is specified is as the stem of the output prefix. So if the output prefix is `path/to/output` then the fasta header will be `>output.max_nest5.min_match7`. Let me know if there are any problems with this convention.
2. Write the `summary.tsv` file to the same directory as all of the other output files (i.e the directory of `--prefix`)

### DIFF
--- a/make_prg/subcommands/prg_from_msa.py
+++ b/make_prg/subcommands/prg_from_msa.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from pathlib import Path
 
 from make_prg import make_prg_from_msa, utils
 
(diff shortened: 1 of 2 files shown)


mbhall88:
Actually, maybe it would be more convenient for downstream analysis to have the header as `>output max_nest=5 min_match=7`?

PR #11, opened 2020-02-12
Viral alignments often have long runs of NNNs in...

rmcolq:
Handle by disallowing alleles in non-match intervals which are entirely N.
The result/consequence is that we need to allow for fewer sequences to be considered after running kmeans_clustering. 
Also add a command line flag for staggered starts, which assumes that dashes at the start of an alignment are due to incomplete sequences rather than true indel differences.


iqbal-lab:
do we want to merge this @rmcolq @leoisl ?

iqbal-lab:
Agree, let's close

Issue #12, opened 2020-04-17
Incomplete ambiguous dna expansion


bricoletc:
I'm delving into `make_prg` some more because looking at interoperability with gramtools, and have found the following behaviour from function `get_interval_seqs`: 

- If you give it the sequence 'RWAAT' it expands to `{"GAAAT", "AAAAT", "GTAAT", "ATAAT"}`
- If you give it the sequence `RRAAT` it expands to `{"GGAAT","AAAAT"}`

The former is complete, the latter is not: missing GAAAT and AGAAT.

(R is ambiguous IUPAC for either G or A)
(Not showing a failing test case because the one I have made is in refactored code)

Is this deliberate?

I am assuming not and have already made a fix which i'll PR in if I get confirmation

rmcolq:
It's not deliberate - feel free to PR and merge in your fix!

PR #13, opened 2020-04-19
Refactoring and more unit tests

bricoletc:
TLDR: shorter and more readable code, and unit tests on the individual functions

My initial main goal was to enforce the following: **the first sequence given in the input MSA should be the first path through the output PRG**

This is currently a strong assumption `gramtools` makes, for genotyping.
I don't know if this codebase was initially designed this way. I did find it was not enforced on master branch. I fixed it and added quite a few tests that give me more confidence it is now enforced. It might be more tests are needed to check it holds for nested prgs later.

In the process, I refactored the code. I made an explicit effort to i) Reduce if/else induced code nesting and ii) Make the code shorter, by for eg removing assertions (with tests) and terser code constructs. I find the code easier to delve into now.

I also added more general unit tests which complement the existing integration tests by testing the individual functions used by `make_prg_from_msa` module.

Also fixes bug #12

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: stable
+    hooks:
+      - id: black
+        language_version: python3.6
--- a/make_prg/__init__.py
+++ b/make_prg/__init__.py
@@ -5,6 +5,6 @@ try:
 except:
     __version__ = "local"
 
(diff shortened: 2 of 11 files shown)


rmcolq:
Sounds excellent. Originally the code was written to preserve first sequence as reference. This has been lost during refactoring because it was not necessary for `pandora`.

bricoletc:
I had a feeling it was, could see several features which looked exactly like that!

PR #16, opened 2020-06-22
add singularity image and update README

mbhall88:
## Added
- Singularity recipe. A built version can be found at <https://cloud.sylabs.io/library/mbhall88/default/make_prg>

## Changed
- Updated README with instructions for using singularity image
- Updated README to reflect current state of project


iqbal-lab:
Suggest modify README to say "for pandora and gramtools"

Issue #17, opened 2020-08-11
Stop producing empty alleles


bricoletc:
`make_prg` can produce sites with i) direct deletions (eg REF AT, ALT "") ii) direct insertions (eg REF "", ALT AT). I refer to REF as the first allele in the site, that's how we embed a 'reference' in gramtools.

Though @mbhall88 has rightly pointed a site made by this tool does not have to translate to one in pandora/gramtools, I argue if we fix this problem here, there's no need to deal with it there. This is especially relevant for gramtools as by default each site produced here is a variant site in the output of `genotype`. It is also important since vcf spec (https://github.com/samtools/hts-specs/blob/master/VCFv4.3.pdf section 1.6.1) states neither REF nor ALT should be empty. 

I will have a look at how to fix this

rmcolq:
So the issue here is to do with multiple levels of nesting which can make it really hard to "get the previous letter and add to both". E.g. suppose that the allele is at level 3, but the bubble its embedded in has no string before the start of the nested split. Gets very messy with the recursion very quickly.

On the other hand, in pandora at least it's pretty easy to fix this so that there are no empty string alleles once the template vcf has been created from the graph and given the vcf reference.

If you can find a fix, great!


Sent from my Samsung Galaxy smartphone.



-------- Original message --------

bricoletc:
Thanks for pointing this out @rmcolq , indeed it looks like we can't easily prepend/postpend sequence to non-match intervals to guarantee the recursive clustering won't eventually hit an empty sequence. However i'd like to at least try to enforce that at 'level 1' we can have that guarantee- WIP!

PR #18, opened 2020-08-12
Dev

bricoletc:
This PR has two main goals:

* Addressing #17 
* Adding more tests and refactoring some of the codebase

## Added

* `IntervalPartitioner` object and associated classes, in interval_partitioner.py
  This object replaces the `partition_alignment_into_intervals` function, which was harder to understand,
  also makes it more easily to modify in order to address #17 and in future.
  Also added more unit tests for interval partitioning.
* This object, when partitioning the consensus string, now spots if a non_match interval induces any empty sequence, and pads any leading match sequence to it. Unit tests to validate the behaviour, and I ran the following toy example:
```
>seq1
AATAATAATAAATTTTGTATAAACTTTACCCTAG
>seq2
AATAATAAT----------------TTACCCTAG
>seq3
AATAATAATAAATTTTATATAAACTTTACCCTAG
```
Before this PR, results in: `AATAATAAT 5 AAATTTTGTATAAACT 6 AAATTTTATATAAACT 6  5 TTACCCTAG` (see the empty third allele`

After this PR, results in :   `AATAATAA 5 TAAATTTTGTATAAACT 6 TAAATTTTATATAAACT 6 T 5 TTACCCTAG` ( the T got prepended)
(max_nesting 5, min_match_len 7)

This does **not** mean #17 is fully resolved, i'll now run on real datasets

## Modified

* An alignment column of all '-' is no longer counted as a match. I cannot find a good reason for it to be **but may be missing something**?
* Moved subcommand parsing to subcommands/ dir, results in less cluttered main()
* get_consensus function shorter and easier to understand + add unit tests for it
* refactored interval partition checks (checking each position of alignment is in one and one only interval,
  and switching non-match to match intervals where needed) into own functions and unit tested them.
* `-v` for verbose switch added at top-level command line

* Overall I strove to make the main module `make_prg_from_msa` smaller and terser, for easier reading/modifying later: went from 499 to 385 lines.


bricoletc:
Hi @mbhall88 @rmcolq , I think this should be merged to master. I can certify that it worked for my gramtools paper analysis, producing PRGs where I found, genotyping on them with gramtools, good concordance with PacBio truth assemblies on two different datasets (TB, Pfalciparum)

make_prg/interval_partition.py
```diff
@@ -0,0 +1,237 @@
...
+        if stop is not None:
+            assert stop >= start
+        self.stop = stop if stop is not None else start
+
+    def modify_by(self, left_delta: int, right_delta: int):
+        self.start += left_delta
+        self.stop += right_delta
+
+    def contains(self, position: int):
+        return self.start <= position <= self.stop
```
mbhall88, line 49: It might be nice to add a docstring mentioning what sort of intervals this class stores. i.e. half-open, open, closed. It seems like this interval is a closed interval?  
Also, just mentioning this incase you aren't aware of it: https://github.com/chaimleib/intervaltree  
I use this for all interval work and it is super nice. I'm not suggesting you replace the class here or anything. Just putting it on your radar if it isn't already in case there is need for it down the line.

make_prg/interval_partition.py
```diff
@@ -0,0 +1,237 @@
...
+class PartitioningError(Exception):
+    pass
+
+
+class IntervalType(Enum):
+    Match = auto()
+    NonMatch = auto()
+
+
+def get_type(letter: str) -> IntervalType:
```
mbhall88, line 22: It might make this a little more readable if this function was added as a static factory constructor on `IntervalType`?

So you would call it something like `from_char` and then when you use it, it is self documenting - `IntervalType.from_char('*')` rather than `get_type('*')` which is a bit more confusing out of context.

Or even try overriding some magic methods to remove the need for a function completely - https://stackoverflow.com/a/55259756/5299417

mbhall88 (approved):
Can't say I understand all the changes, but sounds like your testing has been thorough.

Nice changes from a software engineering point of view. Nice and readable. Should make any future work much easier to do!

make_prg/interval_partition.py
bricoletc: thanks, i added a docstring

make_prg/interval_partition.py
bricoletc: lovely! adding your static constructor suggestion- prefer that than fiddling with magic methods

make_prg/interval_partition.py
```diff
@@ -18,22 +18,24 @@ class IntervalType(Enum):
     Match = auto()
     NonMatch = auto()
 
-
-def get_type(letter: str) -> IntervalType:
-    if letter == "*":
-        return IntervalType.NonMatch
-    else:
-        return IntervalType.Match
+    @classmethod
```
mbhall88, line 21: For future reference @bricoletc this can just be `@staticmethod`. Random example: https://github.com/mbhall88/MemoryUnits/blob/18551958dbf51a9f68e8733ba17b880dbecc0da7/memory_units.py#L45

PR #21, opened 2021-01-22
Add dockerfile

mbhall88:
### Added
- Dockerfile
- `-V` option to get version

### Changed
- A test that was clustering all unique 5-mers was reduced to all 4-mers as the memory
  usage of all 5-mers was causing a segfault when trying to run the tests during the
  docker image build.

### Removed
- Singularity file as it is redundant with the new Dockerfile (that will be hosted on
  quay.io)
- `scipy` dependency. We never actually explicitly use `scipy`.

Once this PR is approved I will set quay.io to build on tagged commits (i.e. version releases)


tests/from_msa/test_cluster_sequences.py
```diff
@@ -383,11 +383,11 @@ def test_GivenAllSequencesSmallEditDist_ReturnsNoClustering(self):
             self.assertEqual(expected_clustering, result)
 
     def test_GivenManyVeryDifferentSequences_EachSeqInOwnCluster(self):
-        # all 1024 distinct DNA 5-mers.
-        # We want each in one cluster, but do not want to run clustering 1023 times
-        all_5mers = list(map("".join, product(standard_bases, repeat=5)))
+        # all 256 distinct DNA 5-mers.
+        # We want each in one cluster, but do not want to run clustering 255 times
+        all_5mers = list(map("".join, product(standard_bases, repeat=4)))
```
bricoletc, line 388: Being picky here, but should be `all_4mers` not `all_5mers`
Also, i realise this comment `# We want each in one cluster, but do not want to run clustering 255 times` is slightly off (my bad), and should be `# We want clustering to keep looking for clusters, and stop at MAX_CLUSTERS (instead of running 255 times)`

tests/from_msa/test_cluster_sequences.py
```diff
@@ -383,11 +383,11 @@ def test_GivenAllSequencesSmallEditDist_ReturnsNoClustering(self):
...
         alignment = make_alignment(all_5mers)
-        result = kmeans_cluster_seqs_in_interval([0, 5], alignment, 5)
+        result = kmeans_cluster_seqs_in_interval([0, 4], alignment, 4)
```
bricoletc, line 390: I'm surprised memory usage on all 5mers was so high, it ran fine on my previous Lenovo laptop. Do you have the log message for this?

(I'm otherwise fine to move down from 5 to 4)

bricoletc (changes requested):
I just have a tiny change request on the test case you modified, but i do think it's important for future code reading. 

All the docker stuff i'm not experienced at all with, so can't comment on, but I think you have the experience with this.

Everything else looks great.

tests/from_msa/test_cluster_sequences.py
bricoletc: Do let me know if youd like me to do this (not sure how though, could I push to mbhall88:master ?)

tests/from_msa/test_cluster_sequences.py
mbhall88: When I profiled it locally is peaked at ~150Mb RAM. Which isn't a lot in the context of a laptop, but is quite high for a test. I guess the reason it was only failing when run in a docker build context is maybe the docker daemon on gets a limited amount of RAM for building?

bricoletc: no idea- but thanks for digging, is worth knowing that was RAM heavy (not very chuffed by that, hope it isn't an issue on large PRGs)

bricoletc (approved):

leoisl:
All looks fine to me, I did not find any issues at all!

leoisl (approved):

PR #26, opened 2021-08-31
Provide some options to reduce output files

mbhall88:
### Added
- `-S`, `--seqid` option to name the PRG sequence, which by default uses the file name.
- `-N` shortcut for max nesting
- `-L` shortcut for min match length
- `--log` to enable specifying log file should go to path. Default behaviour is now that
  log goes to stderr by default
- `--summary` option to specify that user wants a summary file, rather than the default
  behaviour where they get no choice
- `-O`, `--output-type` option to specify what output files are required. Defaults to
  all

### Changed
- Summary file not written by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,11 +110,11 @@ optional arguments:
   -S SEQID, --seqid SEQID
                         Sequence identifier to use for the output sequence/PRG. Default is the file name
   --no_overwrite        Do not replace an existing prg file
-  --summary             Write a summary file
   -O OUTPUT_TYPE, --output-type OUTPUT_TYPE
                         p: PRG, b: Binary, g: GFA, a: All. Combinations are allowed i.e., gb: GFA and Binary. Default: a
   --log LOG             Path to write log to. Default is stderr
 ```
+
 
 ### Nextflow
 
@@ -128,6 +128,7 @@ Requirements: [Nextflow][nf]
 
     Optional arguments:
 ```
+
 
 ## Input
 
@@ -163,3 +164,4 @@ max_forks_make_fasta
 [singularity]: https://sylabs.io/
 [tags]: https://quay.io/repository/iqballab/make_prg?tab=tags
 [conda]: https://conda.io
+
--- a/README.md
+++ b/README.md
@@ -93,22 +93,28 @@ $ make_prg from_msa --help
 usage: make_prg from_msa [options] <MSA input file>
 
 positional arguments:
-  MSA                   Input file: a multiple sequence alignment in supported alignment_format. If not in aligned fasta
-                        alignment_format, use -f to input the alignment_format type
+  MSA                   Input file: a multiple sequence alignment in supported alignment_format. If not in aligned fasta alignment_format, use -f to input
+                        the alignment_format type
 
 optional arguments:
   -h, --help            show this help message and exit
   -f ALIGNMENT_FORMAT, --alignment_format ALIGNMENT_FORMAT
-                        Alignment format of MSA, must be a biopython AlignIO input alignment_format. See
-                        http://biopython.org/wiki/AlignIO. Default: fasta
-  --max_nesting MAX_NESTING
+                        Alignment format of MSA, must be a biopython AlignIO input alignment_format. See http://biopython.org/wiki/AlignIO. Default: fasta
+  -N MAX_NESTING, --max_nesting MAX_NESTING
                         Maximum number of levels to use for nesting. Default: 5
-  --min_match_length MIN_MATCH_LENGTH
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -93,22 +93,28 @@ $ make_prg from_msa --help
 usage: make_prg from_msa [options] <MSA input file>
 
 positional arguments:
-  MSA                   Input file: a multiple sequence alignment in supported alignment_format. If not in aligned fasta
-                        alignment_format, use -f to input the alignment_format type
+  MSA                   Input file: a multiple sequence alignment in supported alignment_format. If not in aligned fasta alignment_format, use -f to input
```
bricoletc, line 96: What does `If not in aligned fasta alignment_format` mean? I think should be `If not in aligned fasta format`?

make_prg/io_utils.py
```diff
@@ -152,13 +152,17 @@ def write_prg(prg_fname: Path, prg_string: str, options: ArgumentParser):
     Writes th prg to `output_file`.
     Writes it as a human readable string, and also as an integer vector
     """
-    with prg_fname.open("w") as prg:
-        header = f">{options.prg_name} max_nest={options.max_nesting} min_match={options.min_match_length}"
-        print(f"{header}\n{prg_string}", file=prg)
+    seqid = options.seqid or options.prg_name
```
bricoletc, line 155: I like this, if I understand it uses `or` shortcircuiting to give priority to seqid :+1:

make_prg/subcommands/from_msa.py
```diff
@@ -67,12 +70,28 @@ def register_parser(subparsers):
...
+    )
     subparser_msa.add_argument(
         "--no_overwrite",
         dest="no_overwrite",
         action="store_true",
-        help="By default, a prg file which already exists is overwritten and the prg is recomputed. Use this option to keep an existing prg file instead.",
+        help="Do not replace an existing prg file",
+    )
+    subparser_msa.add_argument(
+        "--summary", help="Write a summary file", action="store_true"
```
bricoletc, line 85: Do you guys ever use summary file? We could remove it entirely potentially, or just write it out to logging, as its two lines. @leoisl

bricoletc (commented):
Lovely code, I learned from the `OutputType` class, and nice tests on the class.
All the new options make sense to me, I just have a question about `--summary`, see comments.

make_prg/io_utils.py
mbhall88: Correct.

make_prg/subcommands/from_msa.py
mbhall88: I have never used it. Hence why I changed the default behaviour. 

If Leandro never uses it either I vote just get rid of it. It's in the commit history if we ever want it back

README.md
mbhall88: Urgh, yeah that was worded weird. I have made it much more succinct

mbhall88:
> Lovely code, I learned from the `OutputType` class, and nice tests on the class.
> All the new options make sense to me, I just have a question about `--summary`, see comments.

Yeah, I recently implemented something similar in `rasusa` for allowing the user to specify the compression format. I basically stole the idea from `bcftools -O` option

make_prg/subcommands/from_msa.py
leoisl: I've never used it, I vote to remove it also

leoisl (approved):
All fine, I think we all vote to remove the summary file it seems too. But fine for me to merge without this change if you wish

mbhall88:
I guess merging this will require us to bump the version to 0.2.0. Is this going to cause problems with your prototype @leoisl?

leoisl:
Yeah, this is a bit annoying, I am sorry. We can either: bump to version 0.1.2; or just bump to 0.2.0, and treat my fork and this fork totally separate (I don't really like this option).

The issue is that we are still not sure that we will keep the dynamic updates or change to another method to update the PanRG. However, as it is on the paper, I do think we should at least push the dynamic updates to the main repo. I think I will drop all my current priorities, and make the dynamic-update make_prg stable, finishing its interfacing with gramtools, adding unit tests, etc. Then we can merge it to this repo, and therefore have a single main branch of development. The later I push this, the harder will also be to solve the conflicts when merging. What do you think? @iqbal-lab Ok to do that, or would you prefer to leave this merge for later?

iqbal-lab:
Fine with me, I thought it was already merged

mbhall88:
Well it sounds like the best option is to not make a release for the stuff in this PR and instead wait until you merge your fork into this main fork. Then we can release it all as v0.2.0 - a mega release!

Issue #27, opened 2021-10-05
Indels can cause spurious site production


bricoletc:
This is linked to #15 as it leads to production of ambiguous prgs (multiple paths give rise to same sequence)

Clustering code can conclude that there is no meaningful clustering of a set of sequences- e.g. puts each sequence in one cluster.

However the code that [calls clustering](https://github.com/iqbal-lab-org/make_prg/blob/a71f81527c27d38cc75e5aa6bc8a1c8aa69a7c80/make_prg/from_msa/prg_builder.py#L140-L150) can re-run prg-building for sets of sequences that only differ in alignment (i.e. gap positioning), not sequence ([here] (https://github.com/iqbal-lab-org/make_prg/blob/a71f81527c27d38cc75e5aa6bc8a1c8aa69a7c80/make_prg/from_msa/prg_builder.py#L167-L178))

This leads to spurious 'nested variants' and the following pathological example case:
[msp6_ambig.pdf](https://github.com/iqbal-lab-org/make_prg/files/7284578/msp6_ambig.pdf)

Of the 4 paths between nodes labeled 55 and 56, two are identical. This cause gramtools to mis-genotype down the line.

I have a fix that I'm implementing and will PR in

PR #28, opened 2021-10-14
Fix 27

bricoletc:
Here are some changes to produce many fewer graphs with sequence ambiguity- when different paths spell the same sequence, causing problems for downstream read mapping and genotyping.

See #27  for the reduction in ambiguous genotype calls in gramtools across 26 genes in 14 samples, cut out 99% of them on this dataset!

My genotyping performance did not improve overall on this dataset (though it fixed the one in #27), so it may be good to try the changes here on another validation dataset. In fact for upcoming changes by @leoisl we need to do that IMO.

If happy with these changes we should probably version bump.

## Fixed
- Bugfix ([#27][27])- Part 1: clustering termination was not properly detected, causing spurious 
  site production. 
Here we were over-clustering unnecessarily.
- Bugfix ([#27][27])- Part 2: added criteria for not clustering, to reduce construction of 
  ambiguous graphs. 
Here the key change is this:
```
+        num_unique_nongapped = count(
+            remove_duplicates(map(ungap, get_alignment_seqs(sub_alignment)))
+        )
+        num_unique_gapped = count(remove_duplicates(get_alignment_seqs(sub_alignment)))
+        assert num_unique_nongapped <= num_unique_gapped
+        if num_unique_nongapped < num_unique_gapped:
+            return True

```
So, if you have at least one sequence (e.g. `ATTTTTGA`, `num_unique_nongapped` = 1) occurring multiple times in the MSA but as different alignments (e.g. `AT-TTTTGA` and `ATTTT-TGA`, `num_unique_gapped` = 2), we dont do clustering/recursive prg construction. Reason is that the MSA tool did not find a coherent MSA in these cases, and this opens up the door for creating sub-prgs that spell the same sequences.
In our example:
```
AT-TTTTGA
ATTTT-TGA
```
the prg builder could collapse the middle `TT` together, then you have `AT` vs `ATT` in the left sub-prg, and `TTGA` vs `TGA` onthe right sub-prg. And lo and behold, `ATTTTTGA` exists in two different paths.

### DIFF
--- a/make_prg/from_msa/cluster_sequences.py
+++ b/make_prg/from_msa/cluster_sequences.py
@@ -31,10 +31,9 @@ class ClusteringResult(object):
         sequences: Optional[Sequences],
     ):
         mutually_exclusive = (clustered_ids is None) ^ (sequences is None)
-        if not mutually_exclusive:
-            raise ValueError(
-                "Input arguments must be mutually exclusively set to 'None'"
-            )
+        assert (
+            mutually_exclusive
+        ), "Input arguments must be mutually exclusively set to 'None'"
         self.clustered_ids = clustered_ids
         self.sequences = sequences
 
@@ -164,7 +163,9 @@ def merge_sequences(*seqlists: Sequences, first_seq: str) -> ClusteringResult:
             else:
                 result.append(sequence)
     if not first_seq_found:
-        raise ValueError()
+        raise ValueError(
+            "Provided sequence argument not found in provided list of sequences"
+        )
     return ClusteringResult(None, [first_seq] + result)
 
 
@@ -211,7 +212,8 @@ def kmeans_cluster_seqs_in_interval(
 
     num_clusters = 1
     num_sequences = len(seq_to_ids)
-    if num_sequences <= 2:
+    too_few_seqs_to_cluster = num_sequences <= 2
+    if too_few_seqs_to_cluster:
         return merge_sequences(
             seq_to_ids.keys(), small_seq_to_ids.keys(), first_seq=first_sequence
         )
--- a/make_prg/from_msa/prg_builder.py
+++ b/make_prg/from_msa/prg_builder.py
@@ -118,6 +118,12 @@ class PrgBuilder(object):
         min_match_length: int,
         alignment: MSA,
     ) -> bool:
+        """
+        Condition `alignment_has_ambiguity`: if at least one sequence has more than one
+        gapped representation, the aligner could not align the sequence unambiguously.
+        Clustering is then not performed because it can create ambiguous prgs, where
+        different paths spell the same sequence.
+        """
         max_nesting_reached = nesting_level >= max_nesting
         if max_nesting_reached:
             return True
@@ -130,12 +136,15 @@ class PrgBuilder(object):
         num_unique_nongapped = count(
             remove_duplicates(map(ungap, get_alignment_seqs(sub_alignment)))
         )
-        if num_unique_nongapped <= 2:
+        too_few_unique_sequences = num_unique_nongapped <= 2
+        if too_few_unique_sequences:
             return True
 
         num_unique_gapped = count(remove_duplicates(get_alignment_seqs(sub_alignment)))
         assert num_unique_nongapped <= num_unique_gapped
-        if num_unique_nongapped < num_unique_gapped:
+
+        alignment_has_ambiguity = num_unique_nongapped < num_unique_gapped
+        if alignment_has_ambiguity:
             return True
 
         return False
--- a/tests/from_msa/test_cluster_sequences.py
+++ b/tests/from_msa/test_cluster_sequences.py
@@ -35,9 +35,9 @@ class TestClusteringResult(TestCase):
     def test_clustering_result_mutual_exclusivity(self):
         sequences = ["AATA", "AAAA"]
         clustered_ids = [["s0"], ["s1"]]
-        with self.assertRaises(ValueError):
+        with self.assertRaises(AssertionError):
             ClusteringResult(clustered_ids, sequences)
-        with self.assertRaises(ValueError):
+        with self.assertRaises(AssertionError):
             ClusteringResult(None, None)
         ClusteringResult(clustered_ids, None)
         ClusteringResult(None, sequences)


make_prg/from_msa/cluster_sequences.py
```diff
@@ -20,6 +18,33 @@
...
+    returns a set of sequences instead that can be directly used as set of alternative
+    alleles of the variant site under construction.
+    """
+
+    def __init__(
+        self,
+        clustered_ids: Union[ClusteredIDs, None],
+        sequences: Union[Sequences, None],
+    ):
+        mutually_exclusive = (clustered_ids is None) ^ (sequences is None)
```
mbhall88, line 35: This is very clever!!  
However, I had to look up this operator to remind myself which one it is (XOR), and then I had to look up XOR to remind myself of its outcome.  
```suggestion
        mutually_exclusive = (clustered_ids is None) is not (sequences is None)
```
might I suggest this just for clarity's sake?  
Again, what you had was very nice.

make_prg/from_msa/prg_builder.py
```diff
@@ -103,94 +109,128 @@ def get_sub_alignment_by_list_id(
             sub_alignment = sub_alignment[:, interval[0] : interval[1] + 1]
         return sub_alignment
 
+    @classmethod
```
mbhall88, line 112: why is this a classmethod?

make_prg/from_msa/prg_builder.py
```diff
@@ -103,94 +109,128 @@ def get_sub_alignment_by_list_id(
...
+
+        sub_alignment = alignment[:, interval.start : interval.stop + 1]
+        num_unique_nongapped = count(
+            remove_duplicates(map(ungap, get_alignment_seqs(sub_alignment)))
+        )
+        if num_unique_nongapped <= 2:
+            return True
+
+        num_unique_gapped = count(remove_duplicates(get_alignment_seqs(sub_alignment)))
+        assert num_unique_nongapped <= num_unique_gapped
```
mbhall88, line 137: why?

make_prg/from_msa/prg_builder.py
```diff
@@ -103,94 +109,128 @@ def get_sub_alignment_by_list_id(
...
+                    "Clustering did not group any sequences together, each seq is a cluster"
+                )
+                variant_prgs = clustering_result.sequences
+                logging.debug(f"Variant seqs found: {variant_prgs}")
+            else:
+                variant_prgs = self.prg_recur(interval, clustering_result.clustered_ids)
+        assert len(variant_prgs) > 1, "Only have one variant seq"
+        assert len(variant_prgs) == len(
+            list(remove_duplicates(variant_prgs))
+        ), "have repeat variant seqs"
```
mbhall88, line 215: Are we able to suggest a solution here? Having run into some of these assert statements in the past I can attest that they are very confusing and I generally have no idea what to do next

make_prg/seq_utils.py
```diff
@@ -54,14 +54,23 @@ def ungap(seq: str) -> str:
     return seq.replace("-", "")
 
 
-def get_interval_seqs(interval_alignment: MSA):
+def count(iterable) -> int:
```
mbhall88, line 57: is this not just the `len` function?

make_prg/seq_utils.py
```diff
@@ -54,14 +54,23 @@ def ungap(seq: str) -> str:
...
+    return sum(1 for elem in iterable)
+
+
+def get_alignment_seqs(alignment: MSA):
```
mbhall88, line 61:
```suggestion
def get_alignment_seqs(alignment: MSA) -> Generator[str]:
```
will need to import `Generator` from `typing` too

make_prg/seq_utils.py
```diff
@@ -54,14 +54,23 @@ def ungap(seq: str) -> str:
...
+    for record in alignment:
+        yield str(record.seq)
+
+
+def get_expanded_sequences(alignment: MSA):
```
mbhall88, line 66:
```suggestion
def get_expanded_sequences(alignment: MSA) -> List[str]:
```

mbhall88 (approved):
I can't claim to know the correctness of the underlying algorithms, but the test data seems sane.

All my comments are basically just suggestions. Feel free to ignore any you want.

make_prg/from_msa/cluster_sequences.py
bricoletc: Thx. I know what you mean, however I think we should keep this, because i)`is not` is not clearer to me, because what we want to enforce is that they can't both be None, nor both not be None. XOR is the exact definition of that. `is not` is harder to parse mentally ii)The surrounding code documents what's going on, the variable is called `mutually_exclusive` and the raised error if the XOR doesn't hold explains what's being enforced

make_prg/from_msa/prg_builder.py
bricoletc: The reason is i)I want this function to be part of the class `PrgBuilder` but ii)I want to be able to test it in isolation, so it doesn't use any `self` attributes/methods. 

If you have an alternative, I'm open!

make_prg/from_msa/prg_builder.py
bricoletc: Given a sequence in alphabet `[ATCG]`, say you can add gaps to it anywhere `-`. The set of sequences in the alphabet `[ATCG-]` is a superset of that in `[ATCG]`, so that assertion must hold

bricoletc: It's just an extra way to check the counting and ungapping are working fine

make_prg/seq_utils.py
bricoletc: Agreed and done

make_prg/seq_utils.py
bricoletc: Agreed and done

make_prg/seq_utils.py
bricoletc: Good spot. `len` only applies to containers, e.g. a list, not generators, like the `remove_duplicates` function.
`count` is used on generator functions: e.g. prg_builder.py: L130-132. The advantage of using generator functions is they can be chained in a functional programming way, and this avoids building lists!

make_prg/from_msa/prg_builder.py
mbhall88: [`@staticmethod`](https://docs.python.org/3/library/functions.html#staticmethod) is probably what you're after. Although this is just minor minor stuff in this context.

bricoletc: ah yes. the difference is subtle, staticmethod doesn't receive the `class` object as a function parameter. there may be cases where we want the classmethod to be able to use other methods of the class, so i think it's ok to leave as such. but thanks for pointing out i always forget the two exist

PR #33, opened 2021-11-29
make_prg update PR series: 1. make_prg/from_msa refactoring/updates

leoisl:
Starting the series of PRs to the `dev` branch. We first start with `make_prg/from_msa` source code dir.

@bricoletc unfortunately there were two units that I could not easily add to the code in this PR, so I reverted to what was before. This is a shame, as there is development time that was spent to make and tests these units, but adding them would make code readability, understanding and maintenance more complicated, so I decided to not add them. These were the class `ClusteringResult` and the function `merge_sequences`. However, this is just how we represent the results of the clustering. The core logic of the latest PR, which was to not cluster in case of too few unique sequences or alignment ambiguities was incorporated into the code, and will be described in next PRs.

`make_prg/from_msa/prg_builder.py` was moved outside of `make_prg/from_msa`, as it is now used by both the `from_msa` and `update` commands, and it was also heavily refactored due to the need of representing the recursion tree explicitly. Will be included in upcoming PRs, but the changes are such that the two files are not comparable.

`tests/from_msa/__init__.py ` was refactored and moved with a number of other helper functions to `tests/test_helpers.py`.

Next PR will come after this one is approved/merged.


mbhall88 (approved):

make_prg/from_msa/cluster_sequences.py
```diff
@@ -16,32 +16,6 @@
...
-        mutually_exclusive = (clustered_ids is None) ^ (sequences is None)
-        assert (
-            mutually_exclusive
-        ), "Input arguments must be mutually exclusively set to 'None'"
-        self.clustered_ids = clustered_ids
-        self.sequences = sequences
-
-    @property
-    def no_clustering(self):
-        return self.clustered_ids is None
```
bricoletc, line 42: When I wrote it I found this class a little clumsy, but what is your reasoning for removing it? How do you now test if clustering produced a meaningful clustering?

make_prg/from_msa/cluster_sequences.py
```diff
@@ -153,22 +127,6 @@ def extract_clusters(
...
-        for sequence in seqlist:
-            if sequence == first_seq:
-                first_seq_found = True
-            else:
-                result.append(sequence)
-    if not first_seq_found:
-        raise ValueError(
-            "Provided sequence argument not found in provided list of sequences"
-        )
-    return ClusteringResult(None, [first_seq] + result)
```
bricoletc, line 169: This was designed to make sure the first sample ID in the input alignment ends up in the first position of the first cluster- is this behaviour still there after removing this?

tests/from_msa/test_cluster_sequences.py
```diff
@@ -440,33 +433,9 @@ def test_GivenSequencesWithSameKmerCounts_ClusteringInterrupted(self):
...
-        )
-        expected = ["AATAA", "AA", "TT", "TTAAA"]
-        self.assertEqual(result.sequences, expected)
-
-    def test_GivenLongFirst_FirstSeqIsFirstSeqOut(self):
-        result = merge_sequences(
-            self.long_seqs, self.short_seqs, first_seq=self.first_seq
-        )
-        expected = ["AATAA", "TTAAA", "AA", "TT"]
-        self.assertEqual(result.sequences, expected)
```
bricoletc, line 469: These tests were there for a reason!!
The first input sequence in the alignment needs to be the first path in the output prg. gramtools relies on this as the 'reference genome' (carries over the input reference genome sequence in the alignment). Why are these removed??

bricoletc (commented):
A lot of code has been deleted but I've commented places where I don't really understand why- unless subsequence commits to be reviewed maintain the desired behaviour (see comments)

make_prg/from_msa/cluster_sequences.py
bricoletc: The behaviour of this class is different from the `skip_clustering` method in `PrgBuilder` (in `prg_builder.py`). This code is where we've decided not to skip clustering, we do kmeans and for some reason or other (mainly: too few long-enough sequences and as many clusters as sequences) there is no meaningful clustering. In that case we return the sequences directly and they get enumerated as alleles directly. As long as that behaviour is maintained somehow I would OK this change

leoisl: You are right, this class was added back in the last commits, with some changes though. In a part of the new code, we always need the clustering, so the mutual exclusivity of the attributes was removed, and we always know the `clustered_ids` of a `ClusteringResult`. Optionally, if the clustering was not significant, then `sequences` won't be `None`, and we can get the precomputed set of alleles directly. I've added some other trivial methods required by testings and other code. Please recheck the new version

make_prg/from_msa/cluster_sequences.py
leoisl: It was not, but was now added back in the last commits. Please recheck

tests/from_msa/test_cluster_sequences.py
leoisl: Now added back in the last commits. Please recheck

make_prg/from_msa/cluster_sequences.py
leoisl: You will have to see the `Files changed` section in this PR to see this class back

make_prg/from_msa/cluster_sequences.py
```diff
@@ -127,6 +126,69 @@ def extract_clusters(
...
+    result = list()
+    for seqlist in seqlists:
+        for sequence in seqlist:
+            if sequence == first_seq:
+                first_seq_found = True
+            else:
+                result.append(sequence)
+
+    # Note: this is an assert as is it the dev's responsibility to pass the correct first_seq argument (i.e. the user
+    # can't cause this error by erroneous input)
```
bricoletc, line 176: I agree with using an assert instead of throwing error, and with the reasoning, however we could consider not writing that comment, and have this behaviour be a general convention project-wide. But then, for a new dev on this project, we would need a way to tell them this is what we do, and not sure how to do that :')
feel free to ignore this

make_prg/from_msa/cluster_sequences.py
```diff
@@ -162,11 +171,22 @@ def merge_sequences(*seqlists: Sequences, first_seq: str) -> ClusteringResult:
...
-        )
-    return ClusteringResult(None, [first_seq] + result)
+
+    # Note: this is an assert as is it the dev's responsibility to pass the correct first_seq argument (i.e. the user
+    # can't cause this error by erroneous input)
+    assert first_seq_found, f"Provided first sequence argument ({first_seq}) not found in provided list of sequences " \
+                            f"({seqlists})"
+
+    merged_unexpanded_sequences = [first_seq] + result
+    merged_expanded_sequences = SequenceExpander.get_expanded_sequences(merged_unexpanded_sequences)
```
bricoletc, line 181: Not disagreeing with its use, but checking I understand: did you make a `SequenceExpander` class because you're expanding in different ways, eg from a list of sequences versus an MSA object?

It ends up being a bit verbose in my opinion, but I'm not really recommending anything else

make_prg/from_msa/cluster_sequences.py
```diff
@@ -246,18 +263,20 @@ def kmeans_cluster_seqs_in_interval(
 
     no_clustering = num_clusters == 1 or num_clusters == num_sequences
     if no_clustering:
-        return merge_sequences(
+        clustered_ids = [flatten_list(seq_to_ids.values()) + flatten_list(small_seq_to_ids.values())]
+        merged_sequences = merge_sequences(
             seq_to_ids.keys(), small_seq_to_ids.keys(), first_seq=first_sequence
         )
+        return ClusteringResult(clustered_ids, merged_sequences)
```
bricoletc, line 270: Just curious here: `merge_sequences` used to produce a `ClusteringResult` object, why do you prefer not to? Could there be an issue with the sequence ids not matching the sequences?

though if i understand correctly you will only ever have both ids and sequences in `ClusteringResult` when there is no clustering, right? So you would only have one big list of all the ids and one big list of all the sequences.

One option is to have a classmethod in `ClusteringResult` called `make_no_clustering`, with argument the ids, and it does what these lines do

again, feel free to ignore, i just want us, when/if we reread this code in 6 months, to understand quickly what it's doing and also to have a clear guarantee that ids and sequences are matched

bricoletc (approved):
All looks great, I made just a few comments, some are just being curious. Thanks and let's keep going :)

make_prg/from_msa/cluster_sequences.py
bricoletc: actually come to think of this if the first sequence gets moved by `merged_sequences` function, shouldn't the first sequence id be placed first in `clustered_ids`?

make_prg/from_msa/cluster_sequences.py
leoisl: I do agree, and thought about removing the comment also! And I have this comment duplicated at least 3 times in the code base now. I think it might be better to remove the comment, and create a Coding convention wiki page? As you said, this is just a convention, nothing related to the business logic, and it end up polluting the code...

leoisl: it might be better a `coding_convention.md` file in the repo root actually

make_prg/from_msa/cluster_sequences.py
leoisl: It was more for modularisation/refactoring purposes. In `seq_utils.py`, we have 5 variables and 4 functions related just with expanding sequences, so to make this clearer, I wrapped them all in a single `SequenceExpander` class. It is a sort of long source file, with 16 functions, so a bit of modularisation helps. I also thought of refactoring it into its single `sequence_expander.py` source file, but ended up leaving in `seq_utils.py` itself. You will be able to take a better look at it in the next PR, which will have utils-related code

make_prg/from_msa/cluster_sequences.py
bricoletc: @mbhall88 is good with best practices/conventions so i'd like to know his opinion

otherwise i would vote for putting it in wiki page

i wonder if we should have a wiki page or something for the iqbal-group organisation as this is kind of our target audience here

bricoletc: but actually there is no make_prg wiki page right now and perhaps a `coding_convention.md` makes sense.

make_prg/from_msa/cluster_sequences.py
bricoletc: sure!

make_prg/from_msa/cluster_sequences.py
leoisl: > Just curious here: merge_sequences used to produce a ClusteringResult object, why do you prefer not to? Could there be an issue with the sequence ids not matching the sequences?


The main factor for this change is that the previous `merge_sequences()` would return a `ClusteringResult(None, [first_seq] + result)`, i.e. the clustering would be `None`. This is fine for the current code - there is no significant cluster, so let's juts look at the sequences directly. However, in the code of this PR, we explicitly represent the recursion tree, and in some part of this recursion tree, we might have a `MultiClusterNode` (to be seen in next PRs, a node that represent a sub-alignment that will potentially be divided in multiple clusters) that will be split into several `SingleClusterNode`. For the code in the PR, we need to always have the clusters, even if they are not significant, and thus this is why I refactored this function to not return a `ClusteringResult` with no clustering. There is indeed the option to not change this, and keep the return value as `ClusteringResult(None, [first_seq] + result)`, but it is much easier to change this single return value than to change the recursion tree code to accommodate this. The second reason is just for semantics and consistency in the `kmeans_cluster_seqs()` function. We have the `merge_clusters()` functions which merge clusters and return clusters. When I first looked at the `merge_sequences()` function, I thought it would similar: merge sequences and return sequences, but it returned a `ClusteringResult`, which is clusters + sequences. So in `kmeans_cluster_seqs()`, we have a more consistent way to compute the return value: merge clusters, if needed; merge sequences, if needed; and bundle these objects into a `ClusteringResult` object. Please note that I am happy to rollback to the previous code if you think we should, as I think you are the main author of this module.

> though if i understand correctly you will only ever have both ids and sequences in ClusteringResult when there is no clustering, right? So you would only have one big list of all the ids and one big list of all the sequences.

Yes, the recursion tree code always need to know how the clustering was done, even if not significant, to work properly. This is fixable though, we can change it I guess...

> One option is to have a classmethod in ClusteringResult called make_no_clustering, with argument the ids, and it does what these lines do

I'd need further clarifications here, do we keep the old `merge_sequences()`, and add `ClusteringResult.make_no_clustering()`, which is the new `merge_sequences()`?

> again, feel free to ignore, i just want us, when/if we reread this code in 6 months, to understand quickly what it's doing and also to have a clear guarantee that ids and sequences are matched

I don't want to ignore this, this is the crucial part of the review, make sure this looks understandable to both of us and new devs

> actually come to think of this if the first sequence gets moved by merged_sequences function, shouldn't the first sequence id be placed first in clustered_ids?

I think this is yet another bug! Will make a test case for this!

bricoletc: So I think this is the key thing about this potential new bug: it might be better to just order the cluster IDs, and then go from cluster IDs to sequences (using the alignment)

Then you don't need to worry about tracking the two in two separate lines- you see what I mean?

leoisl: So do we rollback to what we had before, just having `clustered_ids` instead of `ClusteringResult`?

bricoletc: I'm actually not sure what to do- i've looked at previous code and there were already two separate functions `merge_sequences` that places first sequence at beginning and `merge_clusters` that places first cluster ID at beginning

Also `merge_sequences` was only used to produced a `ClusteringResult` object corresponding to no clustering anyway, so this 'bug' here might not be one if you don't actually use cluster IDs anyway

leoisl: True, but will solve the bug anyway, as leaving such a bug is just a time bomb, and will increase technical debt, that we should continuously reduce

make_prg/from_msa/cluster_sequences.py
leoisl: I've removed this comment and added a `coding_convention.md` in the project's root, please check

make_prg/from_msa/cluster_sequences.py
leoisl: Fixed this bug and added test cases, please check. Now, in `kmeans_cluster_seqs()`, we always call `merge_clusters()` to merge clusters and to set the id of the first record to the first position of the first cluster, and `merge_sequences()` to merge sequences, if needed, and to set the first sequence in the first position.

make_prg/from_msa/cluster_sequences.py
mbhall88: The convention would be to add this in a `CONTRIBUTING.md` file where we say any assert statement is "<insert the text of that comment>".
I think it's important that if there are any assert statements, they should have a detail description explaining *why* we think this particular event should not be able to happen.

mbhall88:
Ass mentioned in a previous thread, the `coding_convention.md` file added in https://github.com/iqbal-lab-org/make_prg/pull/33/commits/9ce064b88288625139b11060bda232a6b9e75955 should be renamed to `CONTRIBUTING.md` as this is [the conventional file](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) used to describe any dev-specific information about the codebase.

coding_convention.md
```diff
@@ -0,0 +1,11 @@
+This file describes this project's coding conventions. There are many items to add here, but we will start
+small and will improve over time.
```
bricoletc, line 2: I think it's enough to just have `This file describes this project's coding conventions.`

If there are more things to add, they should be there, and if they are not, a dev can't assume they 'should exist'

Sorry for being picky x)

bricoletc:
One last comment + renaming the contribution file to CONTRIBUTING.md and we're good here

make_prg/from_msa/cluster_sequences.py
leoisl: Great, done as proposed!

coding_convention.md
leoisl: You are right, fixed!

PR #34, opened 2021-12-09
make_prg update PR series: 2. make_prg/utils refactoring/updates

leoisl:
Hello,

continuing the update PR series, now moving to `make_prg/utils` module. This module was created in this PR, before we had `seq_utils.py`, `io_utils.py` and `prg_encoder.py` inside the `make_prg` dir. With some refactoring and with quite a good amount of new code, the number of utils files increased from 3 to 7, so I decided to create an utils module. The annoying part of moving + refactoring though is that git sometimes does not cope well with these simultaneous changes, and for some files, e.g. `seq_utils.py`, it says `make_prg/seq_utils.py` was completely removed, and `make_prg/utils/seq_utils.py` is a completely new file, while it was just a file renaming and several changes. This makes reviewing a bit more laborious, but if you prefer we can improve on this (i.e. commit the file renaming operations to this branch, and then PR just the code changes).

Some points to help on review:

* Class `GFA_Output` was extracted from `make_prg/io_utils.py` and refactored into its own source file, `make_prg/utils/gfa.py` with very few changes;
* Many functions in `make_prg/utils/io_utils.py` are marked with a note that they are not unit tested, as it would be a bit complicated to unit test them. However, they are tested by integration tests. I think the only function that remained in this source file from the previous code is `load_alignment_file()`, all the others are new;
* The code in files `make_prg/utils/gfa.py` and `make_prg/utils/recursive_tree_drawer.py` are tested through a single big bang test (i.e. a single test that check if they produce the expected output). This is not ideal, and it is more of a regression test checking if we keep producing the same output over the next versions. However, I think it is fine, as these are not crucial outputs produced by `make_prg`: we don't often look at the `.gfa` files, mostly for debugging. The recursive tree drawer is exclusive for debugging, not for user consumption, to understand the recursive tree generated by `make_prg`, which is rather complicated to visualise in debug mode;
* `prg_encoder.py` has basically no changes.

Thanks a lot for the help!


make_prg/utils/io_utils.py
```diff
@@ -0,0 +1,221 @@
...
+import os
+from loguru import logger
+from make_prg.from_msa import MSA
+from typing import Dict, Optional, List, Tuple
+from collections import defaultdict
+from pathlib import Path
+from zipfile import ZipFile
+
+
+def load_alignment_file(msa_file: [str, Path], alignment_format: str) -> MSA:
```
mbhall88, line 15:
```suggestion
def load_alignment_file(msa_file: Union[str, Path], alignment_format: str) -> MSA:
```

I guess you'll need to add `Union` from `typing`

make_prg/utils/io_utils.py
```diff
@@ -0,0 +1,221 @@
...
+    if not os.path.isdir(path):
+        return
+
+    # remove empty subfolders
+    files = os.listdir(path)
+    if len(files):
+        for f in files:
+            fullpath = os.path.join(path, f)
+            if os.path.isdir(fullpath):
+                remove_empty_folders(fullpath)
```
mbhall88, line 49:
```suggestion
    for f in files:
        fullpath = os.path.join(path, f)
        if os.path.isdir(fullpath):
            remove_empty_folders(fullpath)
```

make_prg/utils/io_utils.py
```diff
@@ -0,0 +1,221 @@
...
+        Path(output_prefix + ".prg.fa").exists()
+        or Path(output_prefix + ".update_DS.zip").exists()
+        or Path(output_prefix + ".prg.bin").exists()
+        or Path(output_prefix + ".prg.bin.zip").exists()
+        or Path(output_prefix + ".prg.gfa").exists()
+        or Path(output_prefix + ".prg.gfa.zip").exists()
+    )
+
+
+def create_temp_dir(output_prefix: str) -> Path:
```
mbhall88, line 69: why use a custom function for this when https://docs.python.org/3.6/library/tempfile.html exists?

make_prg/utils/misc.py
```diff
@@ -0,0 +1,10 @@
+from typing import List
+from itertools import groupby, chain
+
+
+def remove_duplicated_consecutive_elems_from_list(the_list: List) -> List:
```
mbhall88, line 5: `List`s of what? Do you want [`Any`](https://docs.python.org/3/library/typing.html#typing.Any)?

make_prg/utils/msa_aligner.py
```diff
@@ -0,0 +1,137 @@
...
+        return new_sequences_filepath
+
+
+class MAFFT(MSAAligner):
+    @classmethod
+    def get_aligner_name(cls) -> str:
+        return "MAFFT"
+
+    def _prepare_run_tmpdir(self) -> Path:
+        run_tmpdir = self._tmpdir / f"temp.MSA.{uuid.uuid4()}"
```
mbhall88, line 91: Again, why are we creating temporary directories? There's much less chance of things going wrong if we use `tempfile`?

make_prg/utils/seq_utils.py
```diff
@@ -0,0 +1,223 @@
...
+                    f"Allowed bases: {cls.allowed_bases}.\n"
+                    f"Sequence: {sequence}\n"
+                    f"Redo sequence curation.\n"
+                )
+
+    @classmethod
+    def check_all_sequences_are_composed_of_ACGT(cls, sequences: List[str]):
+        for sequence in sequences:
+            sequence_is_composed_of_ACGT_only = set(sequence).issubset(cls.standard_bases)
+            assert sequence_is_composed_of_ACGT_only, f"Sequence ({sequence}) should be composed only of ACTG only."
```
mbhall88, line 106:
```suggestion
            assert sequence_is_composed_of_ACGT_only, f"Sequence ({sequence}) should be composed of ACTG only."
```

mbhall88: I thought we allowed some small number of ambiguities?

make_prg/utils/seq_utils.py
```diff
@@ -0,0 +1,223 @@
...
+    )
+    empty_alignment = len(alignment) == 0
+    if empty_alignment:
+        #  this usually happens if ref or alt are empty, let's check
+        ref_is_empty = len(seq1) == 0
+        alt_is_empty = len(seq2) == 0
+
+        # only one should be empty
+        both_are_empty = ref_is_empty and alt_is_empty
+        assert not both_are_empty
```
mbhall88, line 168: I think future devs will be thankful for a comment here that would indicate a) why this is bad, and b) why you think this should not be allowed to happen.  
Can you speculate as to what might have gone wrong?

mbhall88 (commented):
I will let @bricoletc decide when this is approved. I just have some general questions and comments

make_prg/utils/io_utils.py
leoisl: Yep, thanks for the correction!

make_prg/utils/io_utils.py
leoisl: Thanks again, no idea why I kept this conditional

make_prg/utils/io_utils.py
leoisl: You are right, now using `tempfile.mkdtemp()` (https://github.com/iqbal-lab-org/make_prg/pull/34/commits/ee0a383713c81630fc7505e6d901c07da83d76a6)

make_prg/utils/msa_aligner.py
leoisl: Temporary dirs are needed for three reasons. 1. When we multiprocess, each process create 5 files per locus. To avoid filesystem bloat, we create one dir per locus, and put the 5 files there. These thousands of locus dirs are created inside the temporary dir. It would be too memory heavy to keep all this data in RAM, but maybe we could do some streaming process such that we don't need to write all these files, and later combine them. This might work, but it is a good amount of work just for not creating temp dirs/files, which is not a big issue. 2. `MAFFT` necessarily needs a temp dir (and possibly future aligners). 3. I wanted to control where the temp dir is created i.e. in the output folder, because I see many pipelines/runs failing because the space in the temp/scratch folder is full, especially in clusters. Sure, the standard is to create temp files in the standard temp dir (by default `/tmp`), and users can always override this default with `TMPDIR`, but I think creating temp dirs/files in the output dir will solve some headaches and issues in the future. This is my POV though, would be nice to know if you disagree with any of these points and propose an alternative solution.

mbhall88: I'm not arguing that we don't need temporary directories - we absolutely do. I was just questioning why you are manually creating unique names etc. when `tempfile` handles this exact functionality - albeit thorough tested and supported. You can also tell `tempfile` *where* to create the temporary directories and files - it need not be `TMPDIR`.

Anyway, I'm not suggesting you make drastic changes. I was just flagging this as something that might be worth migrating to in the future and any new temporary file/dir functionality should maybe be handled by `tempfile`. Also, it would mean a lot less code to test.

make_prg/utils/seq_utils.py
leoisl: We do in the input, but not in the output, i.e. PRG. This is done just after expanding the alignments sequences, and when expanding, we transform non-ACGT bases into ACGT, and thus we should have just ACGT in the output. Also, if we encounter a base that is not ACGT, the binary PRG encoder, that produces input for `gramtools` is going to error out, as it just encodes ACGT (I think also both `gramtools` and `pandora` does not handle ambiguous bases in the PRG...).

make_prg/utils/misc.py
leoisl: Thanks, improved the typing of this and the other function in `misc.py` (https://github.com/iqbal-lab-org/make_prg/pull/34/commits/6d7c6524d8f085538ba6db721c305eea388734d8)

make_prg/utils/msa_aligner.py
leoisl: > I'm not arguing that we don't need temporary directories - we absolutely do. I was just questioning why you are manually creating unique names etc. when tempfile handles this exact functionality - albeit thorough tested and supported. You can also tell tempfile where to create the temporary directories and files - it need not be TMPDIR.

Yeah, just found this out when you linked the docs to tempfile, and implemented it here: https://github.com/iqbal-lab-org/make_prg/pull/34/commits/ee0a383713c81630fc7505e6d901c07da83d76a6

> Anyway, I'm not suggesting you make drastic changes. I was just flagging this as something that might be worth migrating to in the future and any new temporary file/dir functionality should maybe be handled by tempfile. Also, it would mean a lot less code to test.

Yep, has already implemented your suggestion here: https://github.com/iqbal-lab-org/make_prg/pull/34/commits/ee0a383713c81630fc7505e6d901c07da83d76a6

make_prg/utils/seq_utils.py
leoisl: Done here: https://github.com/iqbal-lab-org/make_prg/pull/34/commits/8a67d89778a238952037e702a49862be1f318f9b , does it look good? Also realise one of the asserts was not really required...

mbhall88 (approved):
As I mentioned before, I will leave it to Brice to approve this PR as he understands it much better than I do.

make_prg/utils/recursive_tree_drawer.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @staticmethod
+    def _visit(node: RecursiveTreeNode, graph: nx.DiGraph):
+        node_attributes = {}
+        node_is_SingleClusterNode = isinstance(node, SingleClusterNode)
+
+        if node.is_leaf():
+            assert node_is_SingleClusterNode, "Error, a leaf is not a SingleClusterNode"
+            prg_as_list = []
+            node._get_prg(prg_as_list)
```
bricoletc, line 31: I'm guessing there is a reason that `_get_prg` fills in a list passed in as parameter rather than return a constructed list?

bricoletc: (don't need to know it, just that i find the latter clearer)

make_prg/utils/seq_utils.py
```diff
@@ -0,0 +1,220 @@
...
+        Note 1: The sequences are deliberately returned in the order they are received.
+        Note 2: Returned sequences are composed of ACGT only
+        """
+
+        cls.check_if_there_is_sequence_with_disallowed_bases(sequences)
+
+        expanded_seqs = []
+        expanded_set = set()
+        deduplicated_gapless_seqs = remove_duplicates(sequences)
+        for seq in deduplicated_gapless_seqs:
```
bricoletc, line 125: I find it clearer to just do `for seq in remove_duplicates(sequences)`- but this is a minute point, can ignore

bricoletc: although because `remove_duplicates` returns a generator i think it's best not to assign a variable to it- the generator will be exhausted after traversal so the variable becomes useless

make_prg/utils/seq_utils.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def align(seq1: str, seq2: str,
+          match_score: float = 2,
+          mismatch_score: float = -0.9,
+          gap_open_score: float = -1.1,
+          gap_extend_score: float = -1) -> Tuple[str, str]:
+    alignment = pairwise2.align.globalms(
+        seq1, seq2, match_score, mismatch_score, gap_open_score, gap_extend_score, one_alignment_only=True
+    )
```
bricoletc, line 159: I remember @mbhall88  or martin saying BioPython's pairwise2 is v. slow- should we replace it with `edlib`?

make_prg/utils/seq_utils.py
```diff
@@ -0,0 +1,220 @@
...
+        ref_is_empty = len(seq1) == 0
+        alt_is_empty = len(seq2) == 0
+
+        both_are_not_empty = (not ref_is_empty) and (not alt_is_empty)
+        assert not both_are_not_empty, "Alignment error: got an empty alignment from two non-empty input sequences"
+
+        if ref_is_empty:
+            return GAP * len(seq2), seq2
+        else:  # alt_is_empty
+            return seq1, GAP * len(seq1)
```
bricoletc, line 172: when is it meaningful to align an empty sequence with a non-empty sequence? is there a case where, for a prg node(s) to be updated, this can actually happen?

make_prg/utils/seq_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    def get_expanded_sequences_from_MSA(cls, alignment: MSA) -> Sequences:
+        gapless_seqs = list(map(ungap, get_alignment_seqs(alignment)))
+        return cls.get_expanded_sequences(gapless_seqs)
+
+
+def align(seq1: str, seq2: str,
+          match_score: float = 2,
+          mismatch_score: float = -0.9,
+          gap_open_score: float = -1.1,
+          gap_extend_score: float = -1) -> Tuple[str, str]:
```
bricoletc, line 156: I find this function's body's code very confusing, would this work instead
```
ref_is_empty = len(seq1) == 0
alt_is_empty = len(seq2) == 0
assert not (ref_is_empty and alt_is_empty), "Error"
if ref_is_empty:
    return GAP * len(seq2), seq2
elif alt_is_empty:
   return seq1, GAP * len(seq1)
else:
      alignment = pairwise2.align.globalms(
        seq1, seq2, match_score, mismatch_score, gap_open_score, gap_extend_score, one_alignment_only=True
    )
    return alignment[0].seqA, alignment[0].seqB
    
```

make_prg/utils/seq_utils.py
```diff
@@ -0,0 +1,220 @@
...
+        if ref_is_empty:
+            return GAP * len(seq2), seq2
+        else:  # alt_is_empty
+            return seq1, GAP * len(seq1)
+    else:
+        alignment = alignment[0]
+        return alignment.seqA, alignment.seqB
+
+
+def remove_columns_full_of_gaps_from_MSA(alignment: MSA) -> MSA:
```
bricoletc, line 178: Just checking, where is this function used? I'm guessing in later PR code

tests/utils/test_msa_aligner.py
```diff
@@ -0,0 +1,145 @@
...
+
+        popen_mock.assert_called_once_with(args, stderr=subprocess.PIPE, encoding="utf-8", shell=True, env=env)
+        wait_mock.wait.assert_called_once_with()
+
+    @patch("shutil.which", return_value=1)
+    @patch.object(Path, Path.mkdir.__name__)
+    @patch("subprocess.Popen")
+    def test___run_aligner___fail_run(self, popen_mock, *uninteresting_mocks):
+        wait_mock = Mock()
+        wait_mock.wait.return_value = 1
```
bricoletc, line 55: may be worth commenting, or stating in function name, that code 1 signals failure, and that's whats being tested

tests/utils/test_msa_aligner.py
```diff
@@ -0,0 +1,145 @@
...
+            "CCGC",
+            "GGCG",
+            "TTAT"
+
+        }
+
+        expected_updated_alignment = "updated_alignment"
+        actual_updated_alignment = mafft.get_updated_alignment(alignment, new_seqs)
+
+        self.assertEqual(expected_updated_alignment, actual_updated_alignment)
```
bricoletc, line 129: what is actually being tested here? essentially that the execution of a fake mafft runs right?

tests/utils/test_msa_aligner.py
```diff
@@ -0,0 +1,145 @@
...
+    @patch.object(MAFFT, MAFFT._create_new_sequences_file.__name__, return_value=Path("_create_new_sequences_file"))
+    @patch("os.environ", {"fake_key_1": "fake_value_1"})
+    @patch.object(Path, Path.mkdir.__name__)
+    @patch("make_prg.utils.msa_aligner.create_temp_dir", return_value=Path("fake_tmp_dir"))
+    @patch("builtins.open", new_callable=mock_open)
+    @patch.object(SeqIO, SeqIO.write.__name__)
+    @patch.object(MAFFT, MAFFT._run_aligner.__name__)
+    @patch("make_prg.utils.msa_aligner.load_alignment_file", return_value="updated_alignment")
+    @patch.object(MAFFT, MAFFT._cleanup_run.__name__)
+    def test___get_updated_alignment(self, cleanup_run_mock, load_alignment_file_mock, run_aligner_mock, write_mock,
```
bricoletc, line 109: this test is very complicated, and is essentially only testing that the function executes, without checking actual mafft outputs right?

i'm not sure i see this test's value

bricoletc: sorry don't mean to be brutal it looks like it was hard to engineer

bricoletc (approved):
Great work, i didn't spot any bugs/mistakes.

Some code is entirely new and seems to prefigure upcoming update-related code, so I can't verify that it 'works', but I trust all the code i've seen!

I left comments that I think are all at your discretion

make_prg/utils/io_utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+# Note: not unit tested
+def get_stats_on_variants(stats_files: List[Path]) -> Tuple[int, int]:
+    nb_of_variants_successfully_applied = 0
+    nb_of_variants_that_failed_to_be_applied = 0
+    for stat_file in stats_files:
+        with open(stat_file) as stat_file_fh:
+            line_split = stat_file_fh.readline().strip().split()
+            nb_of_variants_successfully_applied_for_this_locus = int(line_split[1])
```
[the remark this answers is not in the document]
leoisl: Just for a slightly higher documentation purposes, so that we know exactly what `int(line_split[1])` means. We could instead remove this variable, and add `int(line_split[1])` to the total, and later, when re-reading the code, we could deduce what this last term means, but I prefer to just be explicit... More documentation never hurts!

make_prg/utils/io_utils.py
```diff
@@ -0,0 +1,219 @@
...
+    Class that represents all files that were generated in processing a PRG (PRG fasta, GFA, pickles, etc)
+    """
+    def __init__(self, PRG: Optional[Path] = None, binary_PRG: Optional[Path] = None,
+                 gfa: Optional[Path] = None, pickle: Optional[Path] = None,
+                 stats: Optional[Path] = None):
+        self.PRG: Optional[Path] = PRG
+        self.binary_PRG: Optional[Path] = binary_PRG
+        self.gfa: Optional[Path] = gfa
+        self.pickle: Optional[Path] = pickle
+        self.stats: Optional[Path] = stats
```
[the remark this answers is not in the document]
leoisl: > I've seen you do this before, specifying the type of the member attribute, but is that really necessary since the type of the function param is already specified?

Yeah, this is just for more documentation purposes, like when I `ctrl` click on the attribute in `pycharm`, and get taken to this line, and immediately see the type of the attribute. Not much necessary though as, as you said, it is also in the function param... Just more documentation!

> Also, because the __init__ is just setting attributes, you could use a dataclass here

Indeed, I was not aware of `dataclass` at all! Will switch to it!

make_prg/utils/io_utils.py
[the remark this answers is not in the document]
bricoletc: yeah it's a balance isn't it- keeping code terse but also documented, i'm happy with your answer ;)

make_prg/utils/io_utils.py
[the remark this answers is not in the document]
leoisl: Switched to `dataclass` in https://github.com/iqbal-lab-org/make_prg/pull/34/commits/963185f294d4daf4d64a5f64abf7ec1c2f827618 . Also managed to switch class `ClusteringResult` to dataclass, as well as 3 other classes in the update code (not shown here). Just loved this std annotation, will make use of it a lot in the future, thanks!

make_prg/utils/io_utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+    @staticmethod
+    def clear_temp_extensions(filename: str) -> str:
+        while True:
+            file_should_be_cleared = filename.endswith(".fa") or filename.endswith(".prg") or \
+                                     filename.endswith(".pickle") or filename.endswith(".stats") or \
+                                     filename.endswith(".bin") or filename.endswith(".gfa")
```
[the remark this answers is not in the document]
leoisl: That is much cleaner, thanks, I indeed added the regexp pattern as a class variable (see https://github.com/iqbal-lab-org/make_prg/pull/34/commits/b12d38d60d132c097dcb34374f2913c9e52d736a)

make_prg/utils/io_utils.py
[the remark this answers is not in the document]
leoisl: Yeah, it is a balance, not sure if there is a better or more correct way to do things in these cases, there is a lot of personal preferences too (maybe we should add an entry about this in `CONTRIBUTING.md`?)

make_prg/utils/io_utils.py
```diff
@@ -0,0 +1,219 @@
...
+                nb_of_variants_successfully_applied_for_this_locus
+            )
+            nb_of_variants_that_failed_to_be_applied_for_this_locus = int(line_split[2])
+            nb_of_variants_that_failed_to_be_applied += (
+                nb_of_variants_that_failed_to_be_applied_for_this_locus
+            )
+    return nb_of_variants_successfully_applied, nb_of_variants_that_failed_to_be_applied
+
+
+# this avoids circular import issues, probably has a better way through this though
```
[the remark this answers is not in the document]
leoisl: I see and agree! Fixed in https://github.com/iqbal-lab-org/make_prg/pull/34/commits/5781e3672ca903a521918b1ba92d5474a84b8131

make_prg/utils/io_utils.py
```diff
@@ -0,0 +1,219 @@
...
+from make_prg import prg_builder
+
+
+# Note: not unit tested
+def create_final_files(temp_dir: Path, output_prefix: str, is_a_single_MSA: bool, output_stats: bool = False):
+    logger.info("Concatenating files from several threads into single final files...")
+
+    logger.info("Creating FASTA file of PRGs...")
+    locus_to_set_of_output_files = SetOutputFiles.get_locus_to_set_of_output_files(temp_dir / "mp_temp")
```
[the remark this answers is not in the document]
leoisl: > so am I right that your new code allows constructing multiple loci in parallel right?

yes

> and then, it concatenates them?

yes

> Does it also make separate PRG files? I would need that for gramtools so that i can then add in the invariant bits of the genome in between
> specifically- i need the .bin files to be kept separate
> i think that's what you do in lines 185-190 below- sorry if i'm making you repeat yourself

Don't worry about it, first of all, I am making a bit messy for you to review this PR, as I made it so large that I had to break into modules, but modules depend on one another... so you have to review the PR with incomplete information, which is far from ideal. Although this code does not make separate PRGs (a single fasta with all PRG strings), we do build separate binary representation of PRGs, that can be used by `gramtools`. As a concrete example, this is the output of an integration test not yet PRed here: https://github.com/leoisl/make_prg/tree/update/tests/integration_tests/data/truth_output/several . If you download `several.prg.bin.zip`, you will see the separate `.bin` files. I think this means all is good for you and `gramtools`? Could you please confirm?

make_prg/utils/msa_aligner.py
```diff
@@ -0,0 +1,124 @@
...
+            raise NotAValidExecutableError(f"Given MSA executable {executable} does not work or is invalid")
+        self._executable: str = executable
+
+    def _set_tmpdir(self, tmpdir: Path):
+        tmpdir.mkdir(parents=True, exist_ok=True)
+        self._tmpdir: Path = tmpdir
+
+    def __init__(self, executable: str, tmpdir: Path = Path("..")):
+        self._set_executable(executable)
+        self._set_tmpdir(tmpdir)
```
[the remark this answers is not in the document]
leoisl: yay! I really like the `get`/`set` pattern to ensure encapsulation, but rarely do it in python... Maybe sth to consider is if we add this pattern to `CONTRIBUTING.md`, i.e., all member variables are private and accessible only through set and get (or property methods). The annoying thing about it is that sometimes the code gets bloated with loads of `set` and `get` functions, so I am unsure about this...

make_prg/utils/seq_utils.py
leoisl: I do agree, leaving an exhausted generator in the code is error prone, implementing your suggestion

leoisl: Done in https://github.com/iqbal-lab-org/make_prg/pull/34/commits/f3bfa155a595c4a8f864d3d74abcf654cdf216b0

make_prg/utils/seq_utils.py
leoisl: Yes, for some specific indels. In rare cases, a PRG node can be an empty string, and if we find a denovo variant through it, we might align an empty sequence (the node's sequence) against a non-empty one (denovo variant). In a more common case, a denovo variant can be the complete deletion of all bases of a PRG node, then we would align the node's sequence with an empty string (this happens somewhat frequently when the PRG node is composed of a single base, e.g. in a SNP, and the denovo varant is the deletion of that base. Actually, in one of the previous iterations of the update feature, this was not implemented, and then it bugged out in real data due to these cases).

make_prg/utils/seq_utils.py
leoisl: Yes, in `RecursiveTreeNode.__init__()`, but I think it might be present in the current code also? It removes columns in the MSA that are full of gaps, like this second column here:

```
A-T
A-C
A-G
```

which can appear when we get sub-alignments from the full MSA.

tests/utils/test_msa_aligner.py
leoisl: Done in https://github.com/iqbal-lab-org/make_prg/pull/34/commits/2ef3adb012a9214ae1f5207e1cf07b4a6a48c09f

tests/utils/test_msa_aligner.py
leoisl: No worries, you were not brutal :p

It basically tests the whole [get_updated_alignment() function](https://github.com/leoisl/make_prg/blob/2e2ae14f74105e3d10f104ab53785fd28fba9309/make_prg/utils/msa_aligner.py#L88-L124), but by mocking heavy function calls, like creating temp dir for MAFFT to run in, creating MAFFT input files, running MAFFT itself, loading the updated alignment and cleaning up the run. This is indeed a complex function to test in the sense that it does lots of heavy stuff. The approach I followed here is mocking all these heavy stuff, as well as calls to other functions that were already unit tested, as I don't think we should require the user to have MAFFT installed in order to test this function, we can just mock it, that is what is usually done when you have a heavy dependency (e.g. a DB connection) during a test. But just mocking is not enough, as if we just mock the MAFFT call, but we invoke it with wrong parameters, the tests still would pass, whereas they should not. We need to make sure that the mocked calls received the parameters the way we expect, and that we know would work in a real execution, and this is asserted [by the end of the test case](https://github.com/leoisl/make_prg/blob/2e2ae14f74105e3d10f104ab53785fd28fba9309/tests/utils/test_msa_aligner.py#L132-L145)

About the value of this test case, IDK actually how to evaluate it, but I think it is worth keeping it...

tests/utils/test_msa_aligner.py
leoisl: Answer in the response of the following comment

make_prg/utils/io_utils.py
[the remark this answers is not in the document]
mbhall88: I don't think that needs to be in the contributing document. If others make contributions they can see the conventions in the code as it is.

make_prg/utils/io_utils.py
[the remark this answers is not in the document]
bricoletc: perfect

make_prg/utils/seq_utils.py
bricoletc: i couldn't find it being used here but it's fine it made sense to me

tests/utils/test_msa_aligner.py
bricoletc: yep works!

bricoletc:
Merge in whenever you're ready @leoisl

make_prg/utils/seq_utils.py
leoisl: Yes, and it made things indeed much clearer, implemented in https://github.com/iqbal-lab-org/make_prg/pull/34/commits/97fe96eba4f294724c4d613718de50eb74f8d3c5 . Thank you!!!

PR #35, opened 2021-12-16
make_prg update PR series: 3. CLI changes

leoisl:
Hello,
this is the 3rd PR, only containing CLI changes. It is the simplest and shortest one to review. My original idea was to leave this one as the last PR (we still have the longest and most complicated one, that contains the whole update code and the explicit representation of the recursion tree), but given that most people have just 2 working days remaining before a long break, I think it is better to just submit this small PR now, and the big one when we come back from the break.

This PR contains just CLI changes for the `make_prg from_msa` command, and the addition of `make_prg update` command.

Main changes, to help reviewing:
* Logging is now more granular (no flags for info, `-v` for debug, `-vv` for trace (this one is only for devs));
* Input/Output changes:
  * Input can now be a single MSA or a dir of MSAs. If it is a single MSA, a single-fasta PRG file, a binary PRG file, a GFA file and a zip containing update structures are created as output. If it is a dir of MSAs, a multi-fasta PRG file, a zip containing all binary PRG files, a zip containing all GFA files, and a zip containing update structures are created as output instead;
  * Replaced `--outdir` by `--output_prefix`;
  * `--prg_name` param removed;
  * `--seqid` param removed;
  * `--no_overwrite` param removed;
  * `--output-type` param removed - we now always build the three types of output files (fasta, GFA, and binary). I am not sure about this one, we can add this param back if you prefer;
* Both commands now can process PRGs in parallel, thus user can specify number of threads to run;
* `make_prg update` subcommand added;

The only tests covering the code in this PR are integration tests, but they are not part of this PR because there are some integration tests that are related to `make_prg update` command, so will choose to PR these tests later, after the update code is reviewed.

Thank you for your help and time!!


make_prg/__main__.py
```diff
@@ -17,23 +19,28 @@ def main():
...
 
     args = parser.parse_args()
 
-    if args.verbose:
-        log_level = logging.DEBUG
-    else:
-        log_level = logging.INFO
-    logging.basicConfig(level=log_level, handlers=[])
+    log_levels = ["INFO", "DEBUG", "TRACE"]
+    log_level = log_levels[min(args.verbose, 2)]
```
mbhall88, line 38:
```suggestion
    log_level = log_levels[min(args.verbose, len(log_levels) - 1)]
```
Slightly less error-prone and no need to remember to change the 2 if we add more log levels

make_prg/__main__.py
```diff
@@ -17,23 +19,28 @@ def main():
...
+    log_file = args.log if args.log else sys.stderr
```
mbhall88, line 39:
```suggestion
    log_file = args.log or sys.stderr
```
This should do the same thing

make_prg/subcommands/from_msa.py
```diff
@@ -1,23 +1,46 @@
...
+        "--input",
         action="store",
         type=str,
-        help="Input file: a multiple sequence alignment",
+        required=True,
+        help=(
+            "Input data. If it is a file, will be read as the supported alignment_format (see "
+            "parameter --alignment_format). A single PRG, GFA and update data structures will "
+            "be created for the given MSA. If it is a dir, will process all files in the dir. "
+            "Multiple PRGs, GFAs and update data structures will be created, one for each MSA"
```
mbhall88, line 26: This is far too verbose for a CLI help menu. This type of information should just be in the docs. Something like the following is sufficient

```suggestion
            "Multiple sequence alignment file or a directory containing such files"
```

make_prg/subcommands/from_msa.py
```diff
@@ -1,23 +1,46 @@
...
+        ),
+    )
+    subparser_msa.add_argument(
+        "-o",
+        "--output_prefix",
```
mbhall88, line 31:
```suggestion
        "--output-prefix",
```
I don't know why, but I have a real dislike for underscores in CLI options. This is the height of fussiness I know, but I get very particular about CLIs. If someone else has a strong opinion *for* underscores, then leave it as is.

If we change this, I guess we have to change the other options with underscores for uniformity.

make_prg/subcommands/from_msa.py
```diff
@@ -1,23 +1,46 @@
...
+        required=True,
+        help="Prefix for the output files",
+    )
+    subparser_msa.add_argument(
+        "-t",
+        "--threads",
+        action="store",
+        type=int,
+        default=1,
+        help="Number of threads (default 1)",
```
mbhall88, line 43:
```suggestion
        help="Number of threads Default: %(default)d",
```

See https://docs.python.org/3.6/library/argparse.html#help

make_prg/subcommands/from_msa.py
```diff
@@ -52,95 +75,94 @@ def register_parser(subparsers):
...
-    )
-    subparser_msa.add_argument(
-        "-S",
-        "--seqid",
-        help="Sequence identifier to use for the output sequence/PRG. Default is the file name",
-    )
-    subparser_msa.add_argument(
-        "--no_overwrite",
-        dest="no_overwrite",
+        "--output_graphs",
```
mbhall88, line 78:
```suggestion
        "--output-graphs",
```
Can you also please add a short option?

make_prg/subcommands/from_msa.py
```diff
@@ -52,95 +75,94 @@ def register_parser(subparsers):
...
+        help="Outputs the recursive tree graphical representation (for development use only)",
     )
-    subparser_msa.add_argument(
-        "-O",
-        "--output-type",
-        help="p: PRG, b: Binary, g: GFA, a: All. Combinations are allowed i.e., gb: GFA and Binary. Default: %(default)s",
-        default="a",
-        type=OutputType,
-    )
-    subparser_msa.add_argument("--log", help="Path to write log to. Default is stderr")
```
mbhall88, line 87: I strongly disagree with this. Is there a functional reason why this is no longer possible? We only recently added this option in.

make_prg/subcommands/from_msa.py
```diff
@@ -52,95 +75,94 @@ def register_parser(subparsers):
...
+        raise FileNotFoundError(f"{input_path} does not exist")
 
-    if options.prg_name is None:
-        options.prg_name = MSA_file.stem
-    ofile_prefix = output_dir / options.prg_name
+    if input_path.is_file():
+        all_files = [input_path]
+    else:
+        all_files = [
+            Path(path).absolute() for path in input_path.iterdir() if path.is_file()
```
mbhall88, line 99:
```suggestion
            path.resolve() for path in input_path.iterdir() if path.is_file()
```

Don't use `Path.absolute` - https://stackoverflow.com/a/44569249/5299417

Also, `iterdir` returns the path as a Path already, so no need to convert

make_prg/subcommands/from_msa.py
```diff
@@ -52,95 +75,94 @@ def register_parser(subparsers):
...
-    logging.info(
-        "Input parameters max_nesting: %d, min_match_length: %d",
-        options.max_nesting,
-        options.min_match_length,
-    )
+def process_MSA(msa_filepath: Path):
+    global options
+    logger.info(f"Generating PRG for {msa_filepath}...")
+    msa_name = msa_filepath.name
+    locus_name = msa_filepath.with_suffix("").name
```
mbhall88, line 108: This implies we don't accept/expect compressed files?

make_prg/subcommands/from_msa.py
```diff
@@ -52,95 +75,94 @@ def register_parser(subparsers):
...
+        gfa.GFA_Output.write_gfa(prefix, prg)
+        builder.serialize(f"{prefix}.pickle")
+        if options.output_graphs:
+            builder.output_debug_graphs(Path(options.output_prefix + "_debug_graphs"))
+
+    except ValueError as value_error:
+        if "No records found in handle" in value_error.args[0]:
+            logger.warning(f"No records found in MSA {msa_filepath}, skipping...")
+        else:
+            raise value_error
```
mbhall88, line 134: Do we know what sort of circumstances would cause this to happen? Maybe we could add a logger.error message instead? Or would the python error be informative?

Sorry, I'mm too used to Rust now where Errors are treated as first-class citizens and are very important.

make_prg/subcommands/from_msa.py
```diff
@@ -52,95 +75,94 @@ def register_parser(subparsers):
...
-            options.MSA,
-            alignment_format=options.alignment_format,
-            max_nesting=options.max_nesting,
-            min_match_length=options.min_match_length,
-        )
-        logging.info(f"Write PRG file to {prg_fname}")
-        io_utils.write_prg(prg_fname, aseq.prg, options)
-        m = aseq.max_nesting_level_reached
-        logging.info(f"Max_nesting_reached\t{m}")
+        logger.info("Multiple files were given as input, multiple outputs will be produced")
```
mbhall88, line 152: This all seems redundant. The user knows whether they passed in a single or multiple files. I say delete all of this

make_prg/subcommands/from_msa.py
```diff
@@ -52,95 +75,94 @@ def register_parser(subparsers):
...
+
+    if io_utils.output_files_already_exist(options.output_prefix):
```
mbhall88, line 154: We should have a `-F, --force` option that overwrites files if they exist

make_prg/subcommands/from_msa.py
```diff
@@ -52,95 +75,94 @@ def register_parser(subparsers):
...
+    output_dir.mkdir(parents=True, exist_ok=True)
+    temp_dir = io_utils.create_temp_dir(output_dir)
+    options.temp_dir = temp_dir
 
-    if options.output_type.gfa:
-        gfa_fname = ofile_prefix.with_suffix(".gfa")
-        logging.info(f"Write GFA file to {gfa_fname}")
-        io_utils.write_gfa(gfa_fname, aseq.prg)
+    logger.info(f"Using {options.threads} threads to generate PRGs...")
+    with multiprocessing.Pool(options.threads, maxtasksperchild=1) as pool:
```
mbhall88, line 163: Could we add some functionality such that if `-t 0` then use use `None`, which in turn uses all available?

make_prg/subcommands/update.py
```diff
@@ -0,0 +1,187 @@
...
+            subparser_update_prg.error(f"{argument} is not a update_DS nor a zip file.")
+        return argument
+    subparser_update_prg.add_argument(
+        "-u",
+        "--update_DS",
+        action="store",
+        type=lambda argument: check_if_is_update_file(argument),
+        required=True,
+        help=(
+            "Filepath to the update data structures (a *.update_DS.zip file created "
```
mbhall88, line 33: I could also be a non-zipped file right?
```suggestion
            "Filepath to the update data structure(s) (*.update_DS.zip or *.update_DS) "
```

make_prg/subcommands/update.py
```diff
@@ -0,0 +1,187 @@
...
+            "by make_prg from_msa)"
```
mbhall88, line 34:
```suggestion
            "from make_prg from_msa)"
```

make_prg/subcommands/update.py
```diff
@@ -0,0 +1,187 @@
...
+        required=True,
+        help="Prefix for the output files",
+    )
+    subparser_update_prg.add_argument(
+        "-t",
+        "--threads",
+        action="store",
+        type=int,
+        default=1,
+        help="Number of threads (default 1)",
```
mbhall88, line 62:
```suggestion
        help="Number of threads Default: %(default)d",
```

make_prg/subcommands/update.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+    subparser_update_prg.add_argument(
+        "--mafft",
+        help="Path to MAFFT executable. By default, it is assumed to be on $PATH",
+        default="mafft",
+    )
+    subparser_update_prg.add_argument(
+        "--output_graphs",
```
mbhall88, line 70: Add short option

make_prg/subcommands/update.py
```diff
@@ -0,0 +1,187 @@
...
+        action="store_true",
+        default=False,
+        help="Outputs the recursive tree graphical representation (for development use only)",
+    )
+    subparser_update_prg.set_defaults(func=run)
+
+    return subparser_update_prg
+
+
+@dataclass
```
mbhall88, line 81: Not sure if you know, but dataclasses were only added in python 3.8. Are we raising our minimum supported python version?

An alternative is to use named tuples like https://github.com/mbhall88/MemoryUnits/blob/18551958dbf51a9f68e8733ba17b880dbecc0da7/memory_units.py#L18

make_prg/subcommands/update.py
```diff
@@ -0,0 +1,187 @@
...
+    aligner: MSAAligner
+
+
+def update(locus_name: str):
+    global options
+    prg_builder_zip_db = PrgBuilderZipDatabase(options.update_DS)
+    prg_builder_zip_db.load()
+    prg_builder_for_locus = prg_builder_zip_db.get_PrgBuilder(locus_name)
+
+    global update_shared_data
```
mbhall88, line 93: There is pretty excessive use of global in this PR. Is this fixed in a future PR?

make_prg/subcommands/update.py
```diff
@@ -0,0 +1,187 @@
...
+    if options.output_graphs:
+        prg_builder_for_locus.output_debug_graphs(Path(options.output_prefix + "_debug_graphs"))
+
+
+def run(cl_options):
+    global options
+    options = cl_options
+    global update_shared_data
+
+    if io_utils.output_files_already_exist(options.output_prefix):
```
mbhall88, line 151: Again, a force option would be ideal

make_prg/subcommands/update.py
```diff
@@ -0,0 +1,187 @@
...
+        raise RuntimeError("One or more output files already exists, aborting run...")
+
+    # read input data
+    logger.info("Checking Multiple Sequence Aligner...")
```
mbhall88, line 155: What does this mean?

make_prg/subcommands/update.py
```diff
@@ -0,0 +1,187 @@
...
+    try:
+        logger.info("Reading update data structures...")
+        prg_builder_zip_db = PrgBuilderZipDatabase(options.update_DS)
+        prg_builder_zip_db.load()
+        logger.info(f"Reading {options.denovo_paths}...")
+        denovo_variants_db = DenovoVariantsDB(options.denovo_paths)
+        update_shared_data = UpdateSharedData(denovo_variants_db, mafft_aligner)
+
+        output_dir = Path(options.output_prefix).parent
+        os.makedirs(output_dir, exist_ok=True)
```
mbhall88, line 173: Why not use the pathlib `mkdir` function?
```suggestion
        output_dir.mkdir(exist_ok=True)
```

mbhall88 (changes requested):

make_prg/__main__.py
leoisl: Agreed

make_prg/__main__.py
leoisl: Agreed

make_prg/subcommands/from_msa.py
leoisl: Agreed

make_prg/subcommands/from_msa.py
leoisl: Will change this and all the other options from scores to dashes for uniformity

make_prg/subcommands/from_msa.py
leoisl: Agreed

make_prg/subcommands/from_msa.py
leoisl: Yes, will add a short option

make_prg/subcommands/from_msa.py
leoisl: There is no functional reason. Will add this back

make_prg/subcommands/from_msa.py
leoisl: Yeah, I have no idea why I converted a `Path` into `Path` again, I think before was some dir iteration using `os` that returned strings, and when I switched to `Path` iteration, forgot to remove the cast. Thanks for the tip on `Path.resolve`!

make_prg/subcommands/from_msa.py
leoisl: You mean input like `gene.fasta.gz`? We don't accept this now, but we can add this feature, won't be hard. Should we do it for this PR, or do we leave a TODO for a next PR?

make_prg/subcommands/update.py
leoisl: Not right now, `make_prg` always creates a `*.update_DS.zip` file, even if you give only one MSA. In this case, this zip file will have a single file inside. This is just for uniformity when receiving a single MSA or a dir of MSAs in the input, `make_prg from_msa` will produce the same file type, that might be input into `make_prg update`. I have a slight preference to leave it like this, as if we create two different file types, we would need more code on output producing, input reading and more tests.

We do this though for the `.gfa` and `.bin` files (i.e. if it is a single MSA, a single `.gfa` and a single `.bin` files are created, but for multiple MSAs these are all bundled together in a single `.gfa.zip` and a single `.bin.zip` files), but we don't have to read these files back at any point later.

If you think this is a bad pattern for a CLI, then I am happy to change. Otherwise, if you have no preference, then I will keep it as is, as I have a slight preference for how it is now

make_prg/subcommands/update.py
leoisl: Agreed

make_prg/subcommands/update.py
leoisl: Oooops, I overlooked this. In the previous PR there was a suggestion to change some classes into `dataclasses`, and I really liked it, and changed a few classes throughout the code to `dataclasses`, but did not pay attention at the minimum python version. I think this is a broader discussion, so I am also inviting @bricoletc here. My opinion is that I would be fine raising our minimum supported python version to `3.7`, as it is fairly standard now, although some systems come preinstalled with only `3.6`. Not sure about raising to `3.8` though, I would like to collect both your opinions, and then we decide.

TODO (not for myself): if we decide to not keep `dataclasses` in order to decrease our minimum supported python version to `3.6` or `3.7`, change the github workflow CI/CD image to use the version we target

make_prg/subcommands/update.py
leoisl: I don't have any intention to fix this in future PRs, because I don't actually know how to fix this without impacting performance, and I prefer to use a few global variables to improve performance. I will make my point clearer.

As far as I know, `python` does not really implement multithreading due to the global interpreter lock - even if you create multiple threads, only one will run at a time. There are ways around this I guess, but will cause more trouble than what it solves, like using non-cpython implementations or trying to disable the GIL, but this can break packages and modules, so I don't think it is worth trying (see https://wiki.python.org/moin/GlobalInterpreterLock). Multithreading would be great, because we would not need global variables, and we could control access to them, but is not an option.

Our only option is thus multiprocessing. We could indeed avoid the global variables by passing them to the target function of `pool.map()` but this incurs into some overhead, as these variable will have to be pickled by the parent process and unpickled by the child process, thus I chose to use this only for light-data variables, which is not the case of `update_shared_data`, which will hold the whole information of all denovo updates that need to be done to the PanRG. However, this data will only be read by the children processes, and never written by them, so if we declare this variable as global, children processes can access this data from the parent process very efficiently due to [linux using Copy-on-write optimisation when forking processes](https://en.wikipedia.org/wiki/Copy-on-write).

In conclusion, I did not find a way to use threads themselves and control access to the variable, so had to forcibly use multiprocessing. And the most efficient way that I found to share heavy data that is created in the parent process with their children is to use a global variable and rely on the Copy-on-write optimisation. I do agree that using global variables is not a good engineering approach and should be always avoided, but I don't see any better option right now, but would be really happy to hear if you know a better solution!

make_prg/subcommands/update.py
leoisl: Agreed

make_prg/subcommands/from_msa.py
leoisl: Done in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/eee33ad6f266062c34a8f0918504cf0cb4a1793e

make_prg/subcommands/from_msa.py
leoisl: Done in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/eee33ad6f266062c34a8f0918504cf0cb4a1793e

make_prg/subcommands/update.py
leoisl: Done in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/eee33ad6f266062c34a8f0918504cf0cb4a1793e

make_prg/subcommands/from_msa.py
mbhall88: Urgh, just saw biopython doesn't accept gzipped fasta. How spread out is our MSA file-reading functionality? Would it only require changing one/two lines? Or is it spread out over multiple modules?

make_prg/subcommands/update.py
mbhall88: Oh sorry, I thought I saw some some code somewhere that was checking for both of those file extensions...

Fine to leave it as it is

make_prg/subcommands/update.py
mbhall88: I'm fine to have 3.8 as minimum. 3.6 is now end of life anyway.

make_prg/subcommands/update.py
mbhall88: Ahhh I see; didn't realise it was a multithreading thing. All good.

make_prg/subcommands/from_msa.py
leoisl: Fixed in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/4e8fdf5f99213aa664e724458bd5de466ee1066d and https://github.com/iqbal-lab-org/make_prg/pull/35/commits/698827c1c6b7200b657587b6af8610ceb71e762f , please check. Also, I've added 6 more integration tests just to ensure we are building the correct output when different `--output-type` parameters are given, but these integration tests are not part of this PR, will be added in a future PR.

make_prg/subcommands/update.py
leoisl: I'm fine to have 3.8 as minimum too now

make_prg/subcommands/update.py
leoisl: It is a log message just before we check if the given `MAFFT` executable runs fine. What do you propose we do here?

make_prg/subcommands/from_msa.py
leoisl: All MSA-file reading functionality are concentrated on [make_prg.io_utils.load_alignment_file](https://github.com/leoisl/make_prg/blob/698827c1c6b7200b657587b6af8610ceb71e762f/make_prg/utils/io_utils.py#L18-L28), which already supports reading `.gz` files. So, we just need to fix other code that was expecting just `.fa` files, which should not be hard, will be done for this PR. Thanks, I totally overlooked this!

make_prg/subcommands/from_msa.py
leoisl: Done in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/f22fdb84dbac5f15708d4a48a240772b58168cd1

make_prg/subcommands/update.py
mbhall88: Maybe just make it debug?

make_prg/subcommands/from_msa.py
mbhall88: Great! Cheers.

make_prg/subcommands/from_msa.py
leoisl: Done in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/647a1071ff246be11284531d092040ce810cba6b in a slightly different manner so that we can print the actual number of threads we are using in the log

make_prg/subcommands/update.py
leoisl: Done in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/33952a681ed44297ef33653631a8a226efef1b29

make_prg/subcommands/from_msa.py
leoisl: Not sure I understand the question... As far as I remember, the only `ValueError` I've got so far is when there is an empty MSA, then `AlignIO.read()` throws a `ValueError` with this specific error message. Just in case we get another type of `ValueError`, we are reraising it to the caller.

I am not sure about changing `logger.warning()` to `logger.error()` as we don't error out on this exception, we simply ignore this MSA and continue execution. The python error/stacktrace is not very informative I think, as we know the circumstances that this exception happens with this error message.

I am not sure I was clear on my answer... Given this, what do you propose us to do here?

make_prg/subcommands/from_msa.py
leoisl: Will be done in next commit!

make_prg/subcommands/update.py
leoisl: Will be done in next commit!

make_prg/subcommands/from_msa.py
leoisl: Done in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/0cda733793f603cd76ad146e62e839bb53844e3d , please check. Also added 5 integration tests to test well this parameter, but these tests will be described only in a future PR

make_prg/subcommands/update.py
leoisl: Done in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/0cda733793f603cd76ad146e62e839bb53844e3d , please check. Also added 5 integration tests to test well this parameter, but these tests will be described only in a future PR

make_prg/subcommands/update.py
```diff
@@ -144,7 +144,7 @@ def run(cl_options):
     options = cl_options
     global update_shared_data
 
-    if io_utils.output_files_already_exist(options.output_prefix):
+    if io_utils.output_files_already_exist(options.force, options.output_type, options.output_prefix):
```
mbhall88, line 147: Sorry to be a pain, but I think the simplest way to do this would have been 
```suggestion
    if io_utils.output_files_already_exist(options.output_type, options.output_prefix) and not options.force:
```
it doesn't really make sense to have force as a parameter to the function

leoisl: I do agree, fixed in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/5d81c9f41d444f5e7bb5b86cd5caf91766147d60 , but I've put the `not options.force` condition first for very slightly better performance. Please check, thanks!

make_prg/subcommands/from_msa.py
leoisl: Sorry for the delay, was on a break. Done in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/4e5492999223c7700502c22b0fb3060894b5953c , please check. Omitted here are two integration tests to ensure `make_prg from_msa` is reading compressed files correctly, will be described in a future PR.

leoisl:
Thanks for all the issues raised @mbhall88 ! I requested a re-review, but please do it at your best convenience, I hope you are enjoying a long break post viva, and will see this just way later! CLI looks much better now, and I have to admit it was the module that I invested the least amount of time and effort, as I was focused mostly on the core update algorithm. But I think most issues were caught by your first review, and it looks much better now, thanks a lot!

make_prg/__main__.py
```diff
@@ -1,8 +1,72 @@
...
+            help="Force overwrite previous output",
+        )
+
+        par.add_argument(
+            "-t",
+            "--threads",
+            action="store",
+            type=int,
+            default=1,
+            help="Number of threads. Default: %(default)d",
```
mbhall88, line 38:
```suggestion
            help="Number of threads. 0 will use all available. Default: %(default)d",
```

mbhall88 (approved):

make_prg/subcommands/update.py
bricoletc: yep i agree though do we say this somewhere e.g. the README?

make_prg/subcommands/from_msa.py
```diff
@@ -32,115 +45,110 @@ def register_parser(subparsers):
...
+
+        if options.output_type.prg:
+            builder.serialize(f"{prefix}.pickle")
+
+        if options.output_graphs:
+            builder.output_debug_graphs(Path(options.output_prefix + "_debug_graphs"))
+
+    except ValueError as value_error:
+        if "No records found in handle" in value_error.args[0]:
+            logger.warning(f"No records found in MSA {msa_filepath}, skipping...")
```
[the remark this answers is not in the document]
mbhall88: I'd probably agree with this. g

make_prg/utils/io_utils.py
```diff
@@ -218,3 +227,11 @@ def create_final_files(temp_dir: Path, output_prefix: str, is_a_single_MSA: bool
...
         output_files.delete_files()
     remove_empty_folders(str(temp_dir))
+
+
+def remove_known_input_extensions(input_filepath: Path) -> Path:
+    if input_filepath.suffix == ".gz":
+        input_filepath = input_filepath.with_suffix("")
+    if input_filepath.suffix == ".fa":
+        input_filepath = input_filepath.with_suffix("")
+    return input_filepath
```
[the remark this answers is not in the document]
mbhall88: I also realised that `fasta` is also a valid extension...

make_prg/subcommands/update.py
```diff
@@ -0,0 +1,184 @@
+import multiprocessing
+from pathlib import Path
+from loguru import logger
+from make_prg.utils import io_utils, gfa
+from make_prg.update.denovo_variants import DenovoVariantsDB
+from make_prg.prg_builder import PrgBuilderZipDatabase, LeafNotFoundException
+from make_prg.utils.msa_aligner import MAFFT, MSAAligner
+from dataclasses import dataclass
```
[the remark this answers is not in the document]
mbhall88: Yep, agreed. You can do this with `isort`. Or if you're using PyCharm, you can run "optimise imports".

leoisl:
Hello,

I've read all your comments and suggestions, and will implement them all, I think they are all correct and worth implementing in this PR. I started with the longest one, which is the refactoring of how we represent the set of output files. This was just done here: https://github.com/iqbal-lab-org/make_prg/pull/35/commits/0d7870a976a97514ee29b1739e06f76bde57766a . Follows some answers and comments about this refactoring to help reviewing:

* A new source file is created, `make_prg/utils/input_output_files.py`, containing the [base abstract class `InputOutputFiles`](https://github.com/iqbal-lab-org/make_prg/pull/35/commits/0d7870a976a97514ee29b1739e06f76bde57766a#diff-2b26fde2e06b2c03fac086eccaea4ee985f84f9869de4382e93936790c29dbe7R14), which represents the set of input files consumed and output files created for each `make_prg` subcommand. This abstract class groups several common methods together, and some [functions from `make_prg/utils/io_utils.py` that help the implementation of this class were moved to this new source file](https://github.com/iqbal-lab-org/make_prg/pull/35/commits/0d7870a976a97514ee29b1739e06f76bde57766a#diff-2b26fde2e06b2c03fac086eccaea4ee985f84f9869de4382e93936790c29dbe7R34-R134).  We also have [a concrete child class `InputOutputFilesFromMSA`](https://github.com/iqbal-lab-org/make_prg/pull/35/commits/0d7870a976a97514ee29b1739e06f76bde57766a#diff-2b26fde2e06b2c03fac086eccaea4ee985f84f9869de4382e93936790c29dbe7R137-R154) for the `make_prg from_msa` command and the [child class `InputOutputFilesUpdate`](https://github.com/iqbal-lab-org/make_prg/pull/35/commits/0d7870a976a97514ee29b1739e06f76bde57766a#diff-2b26fde2e06b2c03fac086eccaea4ee985f84f9869de4382e93936790c29dbe7R157-R164) for the `make_prg update` command;
* Having one class for each `make_prg` command to deal with all the logic of input/output files, with the common stuff abstracted into the base class, made the source files that before managed part of this, `make_prg/subcommands/from_msa.py` and `make_prg/subcommands/update.py ` much simpler, as well as reducing a lot the size of `make_prg/utils/io_utils.py`, which now contains only very general IO util functions;
* Some code duplication were removed, and two functions turned out to be not needed and thus removed, reducing the size of the code and reducing the chance of bugs. Removed functions:
    1. [`clear_temp_extensions`](https://github.com/iqbal-lab-org/make_prg/pull/35/commits/0d7870a976a97514ee29b1739e06f76bde57766a#diff-573e4f99059f5c74029880ec2e3c3e6bed4963e53963f87e33a9d70bd1de077cL109-L116);
    2. [`get_locus_to_set_of_output_files`](https://github.com/iqbal-lab-org/make_prg/pull/35/commits/0d7870a976a97514ee29b1739e06f76bde57766a#diff-573e4f99059f5c74029880ec2e3c3e6bed4963e53963f87e33a9d70bd1de077cL130-L146)
* I hope that understanding this part of the code is not convoluted anymore, now that all the input/output files logic is concentrated in a single class and not scattered around many places;

> I don't have a clear solution. But could we do something like, have the `SetOutputFiles` object have a constructor that takes the parsed `--output-type` string and sets non-`None` values to its attributes accordingly. Then we never use `OutputType` object.

Basically this was done, but I still kept the `OutputType` class. Is this fine or should we necessarily remove it?

> Add locus_name attribute to `SetOutputFiles`. Make one `SetOutputFiles` object per locus to build/update. Each threaded call to from_msa/update uses this object instead- it will know what files to produce.

This was done as said.

> Have a function that takes a list of `SetOutputFiles` and extracts all files. Zip/concatenate from those.

This is done [in the `create_final_files` method](https://github.com/iqbal-lab-org/make_prg/pull/35/commits/0d7870a976a97514ee29b1739e06f76bde57766a#diff-2b26fde2e06b2c03fac086eccaea4ee985f84f9869de4382e93936790c29dbe7R35-R93)

> I have a clear feeling that there is a lot of code redundancy and that something like this approach can reduce it considerably.

Yeah, some code duplication were removed, and two functions turned out to be not needed and thus removed.

Could I have a review on this refactoring please? Will proceed to the other minor comments, but will be able to work on them just on monday.

Thanks a lot for all your help and work, this code is getting much better with your reviews.

bricoletc:
This looks much better to me @leoisl - i like that `InputOutputFiles`' behaviour is now directly coupled to `OutputType` specification. I think it's fine to keep `OutputType` class.

make_prg/utils/io_utils.py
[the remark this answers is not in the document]
leoisl: Yeah, I thought about using regex, but for some reason went for straight `if`s :(
To replace only `.fa`, `.fasta`, `.fa.gz` and `.fasta.gz`, this is the regex I am using in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/65703d97a6f4d4590cbd98abc4b06031527bd16a :  ` re.sub(r"\.(fa|fasta)(\.gz)?$", "", input_filepath)`. Is this fine?

make_prg/subcommands/update.py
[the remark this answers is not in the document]
leoisl: Linting PR will be the last PR. Is this fine or do you want linting in this PR?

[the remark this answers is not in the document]
mbhall88: Oh ok, yes that makes sense I guess. But maybe we need to add something like [this Makefile](https://github.com/snakemake/snakefmt/blob/master/Makefile) to the project so everyone is doing the same thing?

make_prg/utils/io_utils.py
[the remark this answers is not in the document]
mbhall88: Only way to know.....TESTS 😉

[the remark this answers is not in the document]
leoisl: Yes, tests for this function are here : https://github.com/iqbal-lab-org/make_prg/commit/65703d97a6f4d4590cbd98abc4b06031527bd16a#diff-70c27d91365239ff4cb2d6dd1f928fbcfd0585a56068f9ee12424c611017a2d3L18-L57

make_prg/subcommands/update.py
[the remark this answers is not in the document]
leoisl: Yeah, will add a similar Makefile and add an action

make_prg/__main__.py
```diff
@@ -1,8 +1,72 @@
...
+
+def setup_logger(args: argparse.Namespace):
+    if "verbose" in args:  # args.verbose does not exist if make_prg is called with no args
+        log_levels = ["INFO", "DEBUG", "TRACE"]
+        log_level = log_levels[min(args.verbose, len(log_levels) - 1)]
+        log_file = args.log or sys.stderr
+        handlers = [
+            dict(sink=log_file, enqueue=True, level=log_level),
+        ]
+        logger.configure(handlers=handlers)
```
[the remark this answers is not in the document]
leoisl: Great!

make_prg/subcommands/from_msa.py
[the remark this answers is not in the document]
leoisl: I added this just as warning because when running `make_prg` on 37k loci, there were a couple of MSAs that were empty. Or maybe it was another dataset with 10k loci, don't remember exactly... For a single MSA, I do agree erroring out is much better, but I added this as warning thinking on running `make_prg` on a few thousand loci. Do you think that also in this use case, we should error out?

make_prg/subcommands/from_msa.py
```diff
@@ -32,115 +45,110 @@ def register_parser(subparsers):
...
+def run(cl_options):
+    global options
+    options = cl_options
+
+    logger.info("Getting input files...")
+    input_files = get_all_input_files(options.input)
+
+    there_is_no_input_files = len(input_files) == 0
+    if there_is_no_input_files:
+        raise FileNotFoundError(f"No input files found in {options.input}")
```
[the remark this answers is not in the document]
leoisl: Yeah, might be overdocumenting, but it is better than underdocumenting. I followed this pattern through the whole codebase, so I am favouring keeping it as is. Or do you have a heavy preference for simplifying this?

make_prg/subcommands/from_msa.py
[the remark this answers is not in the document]
iqbal-lab: Yes

make_prg/subcommands/from_msa.py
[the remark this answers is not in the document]
bricoletc: nope, this comes down to personal preferences i think, i'm ok with that!

make_prg/subcommands/from_msa.py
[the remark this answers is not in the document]
leoisl: Ok, this is implemented with an integration test to make sure things are working, but I can't push to this PR right now. Just have mobile internet connection, and my provider does not allow to share internet with laptop. Will do it asap.

make_prg/subcommands/from_msa.py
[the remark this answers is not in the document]
leoisl: Alright, thanks for understanding!

make_prg/utils/io_utils.py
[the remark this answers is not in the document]
leoisl: I'd say this is resolved, feel free to reopen in case of any issues.

make_prg/subcommands/update.py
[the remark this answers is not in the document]
leoisl: Closing, feel free to reopen

make_prg/subcommands/update.py
leoisl: Not yet, will add it to dependencies section in Readme, but Readme updates will come in a future PR only.

leoisl: Closing, feel free to reopen

make_prg/subcommands/from_msa.py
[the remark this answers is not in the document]
leoisl: Done in https://github.com/iqbal-lab-org/make_prg/pull/35/commits/3891a60a5af486caebffcd7ac0d3ae7d47792a26 . An integration test was also added, but is not part of this PR, will be described in a future PR.

leoisl:
Hi all again! Thanks for all the comments and corrections. The CLI looks much, much better now, it would not get to this state without your help. I think I addressed all your comments, suggestions and corrections. I am requesting a re-review from @bricoletc to check if all is fine, and will merge once I have his approval.
Cheers

bricoletc:
Looks fine, though I want to point out these comments:

* Reformat the codebase with black
* Add loguru to setup.py dependencies

I understood import refactorings will happen later- will these comments too?

bricoletc (approved):

bricoletc:
I approved the changes but please look at latest comment above before merging- I leave it to you to push those comments and merge, or merge and push those comments later ;)

leoisl:
> Looks fine, though I want to point out these comments:
> 
> * Reformat the codebase with black
> * Add loguru to setup.py dependencies
> 
> I understood import refactorings will happen later- will these comments too? It's a bit weird to have loguru used but not imported as a dependency in this code, but i suppose it's ok to delay adding it to setup.py

Yes, the last PR that will deal with a lot of misc supporting tasks like:
* Code reformatting and linting;
* CI/CD to check for format and linting on next commits to master;
* Dependencies cleanup;
* etc

PR #37, opened 2022-04-19
make_prg update PR series: 5. CLI changes

leoisl:
1. Adds `--suffix` parameter to `make_prg from_msa`. This is in general useful, but it is actually required to use `make_prg` in a `snakemake` pipeline. The main issue I faced is that a `snakemake` pipeline is creating a dir of MSAs to be fed into `make_prg from_msa`, but currently just using the `--input` parameter and giving a dir as input tries to build a PRG from every file in the dir, including the `.snakemake_timestamp` file. This allows us to further filter out which files we want to build a PRG from;
2. Adds `--long-deletion-threshold` parameter to `make_prg update`. Current `make_prg update` implementation has issues dealing with long deletions, this parameter controls the size of the deletion we consider and ignore.

### DIFF
--- a/make_prg/subcommands/from_msa.py
+++ b/make_prg/subcommands/from_msa.py
@@ -29,6 +29,17 @@ def register_parser(subparsers):
         ),
     )
     subparser_msa.add_argument(
+        "-s",
+        "--suffix",
+        action="store",
+        type=str,
+        default="",
+        help=(
+            "If the input parameter (-i, --input) is a directory, then filter for files with this suffix. "
+            "If this parameter is not given, all files in the input directory is considered."
+        ),
+    )
+    subparser_msa.add_argument(
         "-o",
         "--output-prefix",
         dest="output_prefix",
@@ -75,7 +86,7 @@ def register_parser(subparsers):
     return subparser_msa
 
 
-def get_all_input_files(input_path: str) -> List[Path]:
+def get_all_input_files(input_path: str, suffix: str) -> List[Path]:
     input_path = Path(input_path)
     if not input_path.exists():
         raise FileNotFoundError(f"{input_path} does not exist")
@@ -84,7 +95,7 @@ def get_all_input_files(input_path: str) -> List[Path]:
         all_files = [input_path]
     else:
         all_files = [
-            path.resolve() for path in input_path.iterdir() if path.is_file()
+            path.resolve() for path in input_path.iterdir() if path.is_file() and path.name.endswith(suffix)
         ]
     return all_files
 
@@ -132,7 +143,7 @@ def run(cl_options):
     options = cl_options
 
     logger.info("Getting input files...")
-    input_files = get_all_input_files(options.input)
+    input_files = get_all_input_files(options.input, options.suffix)
 
     there_is_no_input_files = len(input_files) == 0
     if there_is_no_input_files:
--- a/make_prg/subcommands/update.py
+++ b/make_prg/subcommands/update.py
@@ -57,7 +57,8 @@ def register_parser(subparsers):
         ),
     )
     subparser_update_prg.add_argument(
-        "--long-deletion-threshold",
+        "-D",
+        "--deletion-threshold",
         dest="long_deletion_threshold",
         action="store",
         type=int,
--- a/make_prg/subcommands/update.py
+++ b/make_prg/subcommands/update.py
@@ -57,6 +57,17 @@ def register_parser(subparsers):
         ),
     )
     subparser_update_prg.add_argument(
+        "--long-deletion-threshold",
+        dest="long_deletion_threshold",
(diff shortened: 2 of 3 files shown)


make_prg/subcommands/from_msa.py
```diff
@@ -28,6 +28,17 @@ def register_parser(subparsers):
             "Multiple sequence alignment file or a directory containing such files"
         ),
     )
+    subparser_msa.add_argument(
```
mbhall88, line 31: Accepting multiple suffixes would be a better approach for completeness?

mbhall88: Could use the same approach that we use in snakefmt with an include/exclude regex? https://github.com/snakemake/snakefmt/blob/d233dfa958479f8ee3b4a341f6ccc3fb8ae9ab78/snakefmt/snakefmt.py#L41-L76

make_prg/subcommands/update.py
```diff
@@ -56,6 +56,17 @@ def check_if_is_update_file(argument: str):
             "Filepath containing denovo sequences. Should point to a denovo_paths.txt file"
         ),
     )
+    subparser_update_prg.add_argument(
+        "--long-deletion-threshold",
```
mbhall88, line 60:
```suggestion
        "-D",
        "--deletion-threshold",
```

make_prg/subcommands/update.py
```diff
@@ -56,6 +56,17 @@ def check_if_is_update_file(argument: str):
...
+        dest="long_deletion_threshold",
+        action="store",
+        type=int,
+        default=10,
```
mbhall88, line 64: This seems a pretty low default? Should we maybe use something a bit bigger?

make_prg/update/denovo_variants.py
```diff
@@ -43,6 +50,11 @@ def _param_checking(start_index_in_linear_path: int, ref: str, alt: str):
         if negative_index_for_variant_pos:
             raise DenovoError(f"Found a negative index for variant pos ({start_index_in_linear_path})")
 
+        deletion_size = len(ref) - len(alt)
```
mbhall88, line 53: So insertions don't cause the same issues?

mbhall88 (changes requested):

make_prg/update/denovo_variants.py
leoisl: Not that I know of. I had a branch that would ignore any indels, but the issues were really caused by deletions only (which makes sense from an algorithmic point of view)

make_prg/subcommands/update.py
leoisl: I had issues with deletions longer than 10, although I don't really remember which values I used, and we ignore few deletions by using 10 as the threshold in paper data. [I increased the log level when we ignore a variant due to a `TooLongDeletion` error](https://github.com/iqbal-lab-org/make_prg/pull/37/commits/699a95993a329c16199d0dbcafabf0cc12701c82), so that the user can easily count how many variants is lost due to this. In the error message, the variant's [delta length is also reported](https://github.com/leoisl/make_prg/blob/699a95993a329c16199d0dbcafabf0cc12701c82/make_prg/update/denovo_variants.py#L56), so a better tuning of this parameter can be done. But I should note it was a single test I did to set this variable default, so definitely 10 is not the ideal value (it is a bit complicated, as the ideal value I think is dataset- and performance-dependent). Anyway, I did not invest much time worrying about this as this is just a temporary solution - once these `make_prg  update` PR series are merged, then I will start working on the next `make_prg` version, which the only objective is to properly deal with indels/long variants (instead of the temporary solution we have here, which breaks these variants into several smaller ones).

Given all this info, please tell me what is your decision: 1) if this is fine as default as is just a temporary solution; or 2) we should improve this value to a better one (which one?) as the version that properly deals with indels/long variants can take some time to be implemented.

make_prg/subcommands/from_msa.py
leoisl: I can't see a better solution than the include/exclude regex, will implement it, thanks!

make_prg/subcommands/update.py
mbhall88: Ok, fine to leave as is

bricoletc (approved):

mbhall88:
@leoisl just wondering if you have an ETA on this (and therefore v0.3)? I am working on drprg again which relies on make_prg, so it would be nice to use the latest conda version rather than the binary from your fork.

make_prg/subcommands/from_msa.py
leoisl: Logged this in an issue: https://github.com/iqbal-lab-org/make_prg/issues/39 and I am moving on

PR #40, opened 2022-11-22
Add sample example and update tests

leoisl:
@mbhall88 sorry for being pre-emptive here, but I am merging this PR directly. This PR just adds the sample example and update `make_prg` tests (code and data). I think it is somewhat unreviewable, with 190 changed files, 2,190 additions and 54 deletions, but it is just the nature of such PR that adds sample example and update tests.

The sample example part is easy to justify why we would need it.

The tests I think is safe to merge because:
1. We have 325 new tests (from 116 to 441 total tests), with test coverage >99% in non-argument parsing code;
2. We have 44 integration tests testing the whole code, 33 for `make_prg from_msa` command and 11 for `make_prg update` command;

Please tell me in the future if I should not do this and all PRs should be peer-reviewed with no exceptions.


iqbal-lab:
this is fine leandro!

mbhall88:
agreed, this is fine

Issue #41, opened 2022-11-22
Recent version of pytest gets stuck on make_prg integration tests


leoisl:
I couldn't find a solution for this for now. As part of the new release, I've tried upgrading python version to 3.10.8 and deps to the latest ones what we could find wheel for it, so that installation is quick. `pytest` unfortunately gets stuck on running make_prg integration tests. In my machine, it always gets stuck at `test___match_nonmatch_shortmatch`. If I remove it, it gets stuck on the next one.

This is not the best idea neither, as Python version 3.7 end-of-life (EOL) is 27 Jun 2023, and we should be compatible with (I guess) the oldest python version still supported. Will revert to the latest python 3.7 version and try to update deps

leoisl:
New release is not compatible with python 3.7:
```
Traceback (most recent call last):
  File "/home/leandro/git/make_prg/make_prg/__main__.py", line 102, in <module>
    main()
  File "/home/leandro/git/make_prg/make_prg/__main__.py", line 96, in main
    args.func(args)
  File "/home/leandro/git/make_prg/make_prg/subcommands/from_msa.py", line 175, in run
    InputOutputFilesFromMSA.create_final_files(successful_input_and_output_files, options.output_prefix)
  File "/home/leandro/git/make_prg/make_prg/utils/input_output_files.py", line 116, in create_final_files
    input_output_file.delete_files()
  File "/home/leandro/git/make_prg/make_prg/utils/input_output_files.py", line 152, in delete_files
    self._delete_file(self.prg)
  File "/home/leandro/git/make_prg/make_prg/utils/input_output_files.py", line 149, in _delete_file
    filepath.unlink(missing_ok=True)
TypeError: unlink() got an unexpected keyword argument 'missing_ok'

Process finished with exit code 1
```

Not hard to fix though...

leoisl:
Neither with 3.8.10, I think this is a reason to fix then

mbhall88:
Seems that parameter was only added in 3.8 (https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink).

I think we can easily revert it to a v3.7 compatible methodology. Juts check if the file exists and if it does, run unlink (the race condition where it has been deleted in the time between those two calls is **extremely** unlikely in our use case)

leoisl:
[...]

PR #42, opened 2022-11-22
make_prg update PR series: 7. Misc changes

leoisl:
This PR makes some unrelated, small misc changes:

1. Removed `.dockerignore` and `Dockerfile` as we will build a `conda` recipe so it is better to make it uniform the way we access `make_prg` through containers using just the biocontainer;
2. Updated the changelog;
3. Updated the readme;
4. Set version to `0.4.0` (before we were thinking on going `1.0.0`, but I think this should now be in some versions, not this one);
5. Now not failing the whole execution if building one PRG was unsuccessful due to a sequence curation error;
6. Fixed minor issues to make `make_prg` compatible with `python 3.7`;
7. Removed `nextflow` script;
8. Added `requirements` file;
9. Updated sample example readme and script;
10. Added portable binary build scripts;
11. Updated `setup.py`;

Sorry for this big list of changes... the ideal would one PR for each, but most changes are trivial


CHANGELOG.md
```diff
@@ -62,8 +98,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   source project CHANGELOG.
 
 
-[Unreleased]: https://github.com/iqbal-lab-org/make_prg/compare/v0.2.0...HEAD
+[Unreleased]: https://github.com/iqbal-lab-org/make_prg/compare/v0.4.0...HEAD
```
mbhall88, line 101:
```suggestion
[Unreleased]: https://github.com/iqbal-lab-org/make_prg/compare/0.4.0...HEAD
```
the tags do not have a v prefix

README.md
```diff
@@ -1,80 +1,78 @@
...
+Multiple Sequence Alignments.
 
 [TOC]: #
 
 ## Table of Contents
+- [Dependencies](#dependencies)
 - [Install](#install)
-  - [Conda](#conda)
-  - [Local](#local)
-  - [Container](#container)
```
mbhall88, line 12: Even though there is a binary available we should still include instructions for other installation ways. In some circumstances I would rather install in conda or a container rather than a binary...

make_prg/__init__.py
```diff
@@ -4,6 +4,6 @@
 try:
     __version__ = get_distribution("make_prg").version
 except:
-    __version__ = "1.0.0"
+    __version__ = "0.4.0"
```
mbhall88, line 7: I don't like this; we now have the version in two places.

See https://github.com/snakemake/snakefmt/blob/master/snakefmt/__init__.py for how we handle version in snakefmt. We use pyproject.toml instead of setup.py, but it *should* work the same (I haven't tested)

setup.py
```diff
@@ -1,25 +1,25 @@
...
 setup(
     name="make_prg",
-    version="0.2.0",
-    description="Build genome graphs (PRGs) from multiple sequence alignments",
-    long_description=readme,
-    long_description_content_type="text/markdown",
+    version="0.4.0",
     packages=find_packages(),
-    url="https://github.com/iqbal-lab-org/make_prg",
+    url="https://github.com/rmcolq/make_prg",
```
mbhall88, line 17: No, we transferred this repo to the lab url https://github.com/iqbal-lab-org/make_prg

mbhall88 (changes requested):
I think we should keep the dockerfile. The biocontainers image will only get build on releases. I find I regularly want a container for a dev commit, which is auto-built on https://quay.io/repository/iqballab/make_prg for every commit.

Also, there is some "packaging" kinks I would like to iron out by converting the setup.py to a pyproject.toml. I'll have a play around today

mbhall88:
Okay I have to do some other work today.
I have pushed a [justfile](https://github.com/casey/just) for running common tasks - should be self explanatory.

One issue I ran into was with pygraphviz. There is a test that uses it and I was struggling to install it on my mac. Do we actually plot anything though?

Tests are failing on python 3.7 when comparing pickle objects and most the integration tests are failing on python 3.10 due to `options` being global I think. The correct solution for that seems to be not using a global... But we will still run into the pickle issue I suspect.

The main change I've made is switching from setup.py to pyproject.toml (via poetry). Not sure how that impacts building the binary but I suspect you might need poetry in the container you build from.

Not sure how familiar you are with it, but the setup is this
1. install poetry https://python-poetry.org/docs/#installation
2. create a conda env with the version of python you want to use for the project (and activate it)
3. change into the make prg directory and run `just install`
4. when you want to run make prg you prefix it with `poetry run` - like so `poetry run make_prg [options]`

I haven't formatted the project yet - I guess we do that when we are happy to merge so it doesn't impact the ability to review code changes.

I added all the dependencies that were explicitly imported in python files in the project - not everything that is listed in the requirements file. And pinned version in as relaxed a  manner as I could.

leoisl:
> I have pushed a [justfile](https://github.com/casey/just) for running common tasks - should be self explanatory.

Thanks, infrastructure is much cleaner now

> One issue I ran into was with pygraphviz. There is a test that uses it and I was struggling to install it on my mac. Do we actually plot anything though?

It is required by `networkx` to draw graphs: https://github.com/iqbal-lab-org/make_prg/blob/7dda6096092016309aa4d7de26ffd52c47c44d33/make_prg/utils/recursive_tree_drawer.py#L53

See also https://networkx.org/documentation/stable/reference/drawing.html#module-networkx.drawing.nx_agraph

I am now getting an error because `networkx` can't draw the graph because `pygraphviz` is not installed.

However, I actually think this functionality should be hidden. This is the parameter related to it:
```
  -g, --output-graphs   Outputs the recursive tree graphical representation
                        (for development use only)
```
This is actually just for me to debug the PRG builder recursive tree, as it is not simple to debug recursion with the common IDE environment. This tree helps a lot, and I will use it to debug for e.g. https://github.com/iqbal-lab-org/make_prg/issues/43 . However, for the user, it has no use, and it requires heavy dependencies: on ubuntu requires `apt` installing `graphviz graphviz-dev`. So I think we should:
1. Keep this feature in the `make_prg` code, because I will use it once in a while;
2. Remove its tests, because it is a simple big bang test, and I will use it just for debug purposes;
3. Remove the parameter associated to it, and this part of `make_prg` code is just invoked if an output env variable (e.g. `make_prg_output_debug_graphs`) is set;

> Tests are failing on python 3.7 when comparing pickle objects

I am fixing this but it requires quite a good amount of change (need to make some recursive `__eq__()` and implement `__hash__()`. Might not be able to finish today.



> most the integration tests are failing on python 3.10 due to `options` being global I think. The correct solution for that seems to be not using a global... But we will still run into the pickle issue I suspect.

Are we making fixing compatibility with 3.10?

> And pinned version in as relaxed a manner as I could.

My tests don't run anymore, I get stuck just like in https://github.com/iqbal-lab-org/make_prg/issues/41

make_prg/__main__.py
```diff
@@ -38,15 +38,6 @@ def setup_common_last_options(parsers: List[argparse.ArgumentParser]):
...
 
-        par.add_argument(
-            "-g",
-            "--output-graphs",
-            dest="output_graphs",
-            action="store_true",
-            default=False,
-            help="Outputs the recursive tree graphical representation (for development use only)"
-        )
-
```
mbhall88, line 49: You could just make this a hidden arg? https://stackoverflow.com/questions/11114589/creating-hidden-arguments-with-python-argparse
Saves having to do all the nother stuff associated with this commit? Although seems like it'll do the same thing, so leave it as is as undoing this commit will create more work

leoisl: Yeah, I didn't know that. I think it can be implemented with both hidden args and env variables. Env variable makes some stuff easier, like [this method](https://github.com/leoisl/make_prg/blob/update_PRs_2/make_prg/utils/misc.py#L22-L23) to check if we should output debug graphs, that is used in some places in the code without needing to pass the argument over and over. I think we can keep like this for now, and maybe create an issue to refactor it?

leoisl:
Summary of changes I did today:
* Downgraded scikit-learn to `0.24.2` otherwise I can't run tests (https://github.com/iqbal-lab-org/make_prg/issues/44);
* Some refactoring (https://github.com/iqbal-lab-org/make_prg/pull/42/commits/3233aa696ddb67fa0129d9608f81c724c1635cef and https://github.com/iqbal-lab-org/make_prg/pull/42/commits/62126e66ea56430ebaee9efeb6e051247e3d81ba);
* Tried to fix integration tests (https://github.com/iqbal-lab-org/make_prg/pull/42/commits/3acbdbf0a0a0c10bb2c6d8f246360780d195b81d and https://github.com/iqbal-lab-org/make_prg/pull/42/commits/0fa0a33da0b4b37744d6bf4d3b655506bdfd6c4b);
* We just produce debug graphs if we set an env variable and also installs extra poetry deps (https://github.com/iqbal-lab-org/make_prg/pull/42/commits/988eb4a308dc72b75c1ab4da6048fbedcdacedba , https://github.com/iqbal-lab-org/make_prg/pull/42/commits/164812f1c80b9db9b1f3252726af7dbb82842767 and https://github.com/iqbal-lab-org/make_prg/pull/42/commits/e5f64a26541f750a14299d8ee8208a489f211195)

make_prg/__main__.py
mbhall88: I think just leave it as it is. As long as we (you) know how to generate the trees that's all that matters. Don't worry about an issue for it

leoisl:
TODO tomorrow:

- [ ] Add unit tests for the new code of https://github.com/iqbal-lab-org/make_prg/commit/3acbdbf0a0a0c10bb2c6d8f246360780d195b81d;
- [ ] Two tests not working on Michael's env: `Test_From_MSA_Integration_Full_Builds.test___several___compressed_and_uncompressed_input_files` and `Test_From_MSA_Integration_Full_Builds.test___several_alignments`

mbhall88:
So the changes I've made all relate to tooling/packaging.

We now support python versions 3.8+ and test for these on the CI (I've tested all four versions locally and they seem to work (apart from the two failing tests))

I have pinned scikit-learnt o `^1.0` again as I was only getting the test hang locally on version 3.7 and I got some weird CPython errors using 0.24.2 on python 3.10 on both mac and linux.

Annoyingly I can only debug the CI workflow once this all gets merged to master. But I've copied mostly from existing workflows I have so it should be fine.

I've also removed the use of `global` variables and instead explicitly pass those variables around as needed (see https://github.com/iqbal-lab-org/make_prg/pull/42/commits/f03be8ec34ce32659243cf1d96f450b9f194570b)

You can also pass pytest options through `just`. So for verbose testing, you can do `just test -v`

leoisl:
So in the previous commits I fixed the tests, all is passing now. However, `scikit-learn = "~1.0"` is still getting my tests hanging, so I downgraded to `0.24.2`. After doing this in the [CI branch](https://github.com/leoisl/make_prg/tree/CI), [everything runs fine from python 3.7 to python3.11](https://github.com/leoisl/make_prg/actions/runs/3540860485). On ubuntu runners, tests run in <10 seconds. On os x runners, it takes 1.5 minutes to run, still fine. However, it takes ~15m to install the project, which is weird but ok. In the end what we have is ubuntu runners completing CI in ~1m and os x runners in ~20 min. Then, in the [scikit_1.0 branch](https://github.com/leoisl/make_prg/tree/scikit_1.0), I removed the support for python 3.7 and upgrade scikit-learn to "~1.0". However, all ubuntu runners hanged on CI: https://github.com/leoisl/make_prg/actions/runs/3541011094 , I cancelled the workflow after waiting for 1 hour. The mac os runners ran just fine with similar runtime as using `scikit-learn = "0.24.2"`. So I think with this we should stick to `scikit-learn = "0.24.2"`.

leoisl:
Well, latest commits was just me fiddling with CI trying to upload the coverage to this PR, but I think I will postpone this and work on the things we need to finish this PR/release. Just two important points:
1. We exclude some files when computing coverage, explanation here: https://github.com/iqbal-lab-org/make_prg/pull/42/commits/a5f7429e98e7f563ece99bf91be863da48f78f57
2. CI fails if cov < 98%: https://github.com/iqbal-lab-org/make_prg/pull/42/commits/af2b175d6383d1307a2f45102a34fe1ba2b5ef6d

leoisl:
Will finish last tests tomorrow, will bring coverage to 99%+

mbhall88:
Awesome work!! Happy with everything.

I added a `coverage` recipe to the just file that will open a html report in the browser. Also added a `tag` recipe which will gived you the git commands to run for tagging and pushing a release from git CLI.

I also removed the `-vv` default for pytest as it was killing me. I added `-vv` for the CI though, and you can obviously do it when running `just test -vv`

mbhall88 (approved):
All good by me now. One small reminder is to make sure you tag the version as `0.4.0` instead of `v0.4.0`.

We can reuse some of the release pipeline from [here](https://github.com/mbhall88/tbpore/blob/main/.github/workflows/release.yaml) too. Although, looks like [make prg has been distributed on pypi](https://pypi.org/project/make-prg/) by @iqbal-lab pypi account? We will need to [generate an API token](https://pypi.org/help/#apitoken) (@iqbal-lab might need to do that unless anyone else has ther pypi login) to add to this repo as a [github secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets) (any of us can do that bit) for the auto release to work

leoisl:
Thanks! I still need one or a couple of hours to finish all tests regarding the new and changed code, and to fix https://github.com/iqbal-lab-org/make_prg/issues/43 . Will add the latter as an integration test as well. If there are no more issues, we will have the release then by early next week.

leoisl:
Recent commits summary:
* Unit tests finished, coverage is now 99.73%;
* Added `pyinstaller` to a `poetry` extra section (`precompiled_binary`). For this to work, I had to set `python` version from `python = "^3.8"` to `python = ">=3.8,<=3.11"`. I hope this is fine @mbhall88 - we are losing `python 3.12`, but just realised that is not even released yet, still under dev;
* Updated `Dockerfile` with a workaround to make `pyinstaller` work;
* Simplified a lot the precompiled binary scripts: it basically uses the container built with the `Dockerfile` and just invokes `pyinstaller`;
* Removed `requirements.txt`;

Please tell me if you find any issues @mbhall88

leoisl:
Tomorrow I will fix the last outstanding issues:
* Move `update_shared_data` to a singleton object;
* https://github.com/iqbal-lab-org/make_prg/issues/43;

And then I think we are ready to merge and release

mbhall88:
All changes are fine by me!

leoisl:
I am merging this and submitting new PRs for the outstanding issues, will be better for reviewing

PR #46, opened 2022-11-29
Implements update_shared_data as a class attribute in a singleton pattern and drops Mac OS X support

### LINKED ISSUE #45: Global variables, multiprocessing and pickling
Hey @mbhall88 , I finally remembered the reason for me to use global variables for the CLI `options` and for the `update_shared_data` variable, which holds a [DenovoVariantsDB object](https://github.com/leoisl/make_prg/blob/4d83c8efcbe2c7f6c89a713955b084529bfbe414/make_prg/update/denovo_variants.py#L293-L297), that represents all denovo variants for a run, a super heavy object. The reason was all about performance. When we do multiprocessing in python, one of the steps is pickling the input args on the parent process and unpickling them later on the child process to run the specified function. The issue is that pickling/unpickling involves disk operations and is slow.

However, I think removing the CLI `options` from global is the right thing to do. We are already pickling/unpickling a variable when running the multiprocessing part. This `options` variable is very light, and is probably fast to pickle/unpickle, although we have to do it thousands of times.

On the other hand, [I don't think the `update_shared_data` variable should pickled/unpickled everytime we run an update](https://github.com/leoisl/make_prg/blob/4d83c8efcbe2c7f6c89a713955b084529bfbe414/make_prg/subcommands/update.py#L188-L194). This data structure is huge and will definitely kill performance. We will spend most of our time pickling/unpickling this object. The trick here was to set this variable as global, so it would exist in the parent process, and would be copied to the child processes when the fork happened when multiprocessing. Although this does not seem we gain efficiency, because we would still copy this heavy data structure to the child process, linux (and I hope MacOS) implement the [Copy-on-write optimisation](https://en.wikipedia.org/wiki/Copy-on-write): it just copies the pages from the parent to the child process if the child makes a write to the page. In our case, the `update_shared_data` variable is read-only from the child, i.e. the child does not change anything, so in our case it is never going to copy the page from the parent to the child process, allowing for effective memory sharing between the parent and the child. I actually had this issue ~1.5y ago: https://github.com/iqbal-lab/pandora1_paper/issues/252#issuecomment-813302128 , which is when I chose to move to the global variable approach. My mistake was not to document this.

So, in summary, the idea of multiprocessing in python is to gain performance, but if we have input/output args that are huge, we have to pickle/unpickle them, and we lose all the performance we wanted to gain. If the child processes never change the input args, then we can efficiently share the args in memory using global variables. Global variables are bad though, so we should use a singleton approach. I think to be sure we should measure the performance on the 20-way using the current code, which will pickle/unpickle the heavy data structure, and then try the singleton approach...

leoisl:
Sorry that this PR is now doing two things.

This PR firstly describes the implementation of `update_shared_data` as a class attribute in a singleton pattern. This has the benefit of sharing in-memory data efficiently between the parent and child processes when multiprocessing during `make_prg update` operation, without the drawbacks of using `global`. Due to how singleton objects work, the update integration tests now have to run on different subprocesses, so `pytest-forked = "^1.4.0"` has to be added to `dev.dependencies`. Closes #45 

Secondly, it drops support for `Mac OS X`. There are several reasons:
1. Our tools that depend on `make_prg` are mostly tested on `linux`, e.g. `make_prg` and `pandora` provides precompiled binaries just for `linux`, and sample example use these (easy to change though);
2. `pandora` is just available on linux through `conda`;
3. `pandora` does not implement multithreading in `Mac OS X`;
4. next release of `pandora` has not been tested on `Mac OS X`;
5. `drprg` is tested just on linux runners;
6. See slack for more discussion on this;

This support drop can be just temporary. Now we can proceed and finish the new `make_prg` and `pandora` releases this week, enabling `drprg` to use these new releases. We can discuss further, and if we decide to indeed support `Mac OS x`, we can work on this support next week.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         python-version: [ 3.8, 3.9, '3.10', '3.11' ]
-        os: [ ubuntu-latest, macos-latest ]
+        os: [ ubuntu-latest ]
     defaults:
       run:
         shell: bash -el {0}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,9 @@ operations.
 - Several minor bugs;
 - Heavy refactoring of almost the whole codebase;
 
+### Removed
+- Support for `Mac OS X`;
+
 ## [0.2.0] - 2021-10-27
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -3,6 +3,11 @@
 A tool to create and update PRGs for input to [Pandora][pandora] and [Gramtools][gramtools] from a set of 
 Multiple Sequence Alignments.
 
+# Support
+
+We fully support `make_prg` on `linux`.
+No guarantees are made for other operating systems.
+
 [TOC]: #
 
 ## Table of Contents
--- a/justfile
+++ b/justfile
@@ -23,10 +23,6 @@ fmt:
 install:
     poetry install
 
-# installs all extra packages
-install_full:
-    poetry install --extras "debug_graphs"
-
 install-ci: install-mafft
     poetry install --no-interaction
 
--- a/make_prg/subcommands/update.py
+++ b/make_prg/subcommands/update.py
@@ -1,15 +1,15 @@
 import multiprocessing
-from dataclasses import dataclass
 from pathlib import Path
 
 from loguru import logger
 
 from make_prg.prg_builder import LeafNotFoundException, PrgBuilderZipDatabase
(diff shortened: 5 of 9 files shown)


leoisl:
Update integration tests works on linux, but does not work on `mac os x`... trying to understand why and fix

leoisl:
Sharing singleton objects does not seem to work on `os x`: https://github.com/leoisl/make_prg/actions/runs/3574053921/jobs/6008850030#step:8:549

leoisl:
Now that `Mac OS X` support has been dropped, CI works, ready for review

mbhall88 (approved):

PR #48, opened 2022-11-30
Changing the node building priority in NodeFactory.build()

### LINKED ISSUE #43: Understand why latest executable of new make_prg version did not produce the expected graph
Related to https://github.com/rmcolq/pandora/issues/310#issuecomment-1322417386

We solve this issue and I think we are ready to make the new release.

leoisl:
This PR makes it explicit the priority of which node we build in `NodeFactory.build()`. This is important to prioritise which types of nodes we try to build first. This priority directly translates to how we partition the MSA in the recursive clustering and collapse algorithm, so it is at the core of the `make_prg` algorithm. In summary, we try to first build a leaf if we can, as if we can build a leaf, there is no need to recursively partition the MSA horizontally or vertically. If that does not work, we try to build a multi interval node, i.e. try to partition the MSA vertically. If that does not work, we try to build a multi cluster node, i.e. try to partition the MSA horizontally. If that does not work, then it means we can't partition the MSA anymore, then we force build a leaf. We have a corner case when building the root, which is slightly different. In more details this is the node building priority we apply:

```
Node building priority depends on the type of node we are building:
1. If type of node is root
    1.1. Try to build a leaf
    1.2. Force build a multi interval node (even if we have just a single mismatch interval)
2. If type of node is non-root
    2.1. Try to build a leaf
    2.2. Try to build a multi interval node
    2.3. Try to build a multi cluster node
    2.4. If all fails, force build a leaf
```

Several unit tests were added, and two integration tests validating the sample example were added.

Closes #43


mbhall88 (approved):

iqbal-lab:
does this explain that SNP thing @leoisl ?

leoisl:
yep! although it would be interesting to check the impact of this change on the 20-way dataset... I think we should test new make_prg + new pandora releases on both 20-way and drprg...

PR #52, opened 2022-12-01
Embed mafft into make_prg

### LINKED ISSUE #50: MAFFT is portable for linux and we can embed it into make_prg
See https://mafft.cbrc.jp/alignment/software/linuxportable.html

Like this, `make_prg` dependencies are all `python` deps, and there is no need to install `mafft` anymore

leoisl:
Now we have no non-python dependencies anymore. We don't need to have `mafft` to run update integration tests, nor to run the sample example. The precompiled binary and installation through `pip` are now self contained. `conda` and containers don't need to install `mafft` anymore, but that is a very minor improvement to these installation processes, as it is quite easy to install `mafft` in a `conda` or in a container. Removed `--mafft` parameter from `make_prg update` subcommand.

Closes #50


mbhall88 (approved):
I am going to be a pessimist and say I think this change will cause more problems than it solves.

iqbal-lab:
But you approve pessimistically? Or not approve?

leoisl:
I leave to both of you the decision to accept or reject this PR! Any decision is totally fine by me!

mbhall88:
I approve pessimistically 😆

leoisl:
I think this means I can merge and release, doing it!

PR #60, opened 2023-07-11
Properly handling Ns in make_prg

leoisl:
Our handling of `N` bases  in `make_prg` has not been appropriate I guess. An extreme example comes from @Danderson123 . He has an MSA with 789 alleles, and just a single allele has two runs of 10 `N`s. At some point during the recursive cluster and collapse algorithm, we get this part of the allele isolated, and we skip this gene:
```
2023-07-11 10:09:41.175 | WARNING  | make_prg.subcommands.from_msa:process_MSA:148 - Skipping building PRG for alsB. Error: All sequences in this slice contained N. Redo sequence curation.
Sequences: ['GAANNNNNNNNNN']
```

I think this might be worst than erroring out, specially in cases where we build PRGs from thousands of genes, this just gets buried in the log files, even though it is a `WARNING` log message...

The rationale to refuse to build the PRG from MSAs where we have full columns with `N` might be good, but the recursive nature of this problem makes it hard to generalise (e.g. we should refuse building a graph from a 10-allele MSA if some columns are all `N`s, but if this was a 1000-allele MSA and just these had 10 had the `N`s, it would be ok...). Anyway, we chatted and it seemed to us the best solution would be to replace `N` with the consensus of each column. If the whole column is N and/or gaps, we replace it by a random base, with the seed initialised with the alignment sequences, i.e. this replacement is reproducible if the same alignment is given. This is done upfront when loading the MSA so that we take the whole MSA as the context to find the consensus for each column.

This is a small PR, the huge number of additions comes from the integration tests I added (I used the 3 genes @Danderson123 pointed out in his minimum working example).

This should fix https://github.com/rmcolq/pandora/issues/333


make_prg/utils/seq_utils.py
```diff
@@ -234,3 +237,46 @@ def get_consensus_from_MSA(alignment: MSA) -> str:
...
+            record.seq[i] for record in alignment if record.seq[i] != "-"
+        )
+
+        # If there are no residues other than gaps at this position, use a random base
+        if len(pos_counts) == 0:
+            consensus += random.choice("ACGT")
+            continue
+
+        # Find the residue with the highest count
+        max_residue, max_count = pos_counts.most_common(1)[0]
```
mbhall88, line 267: What about the scenario where the two most common have the same count? Might be better to randomly select? The order in the `Counter` will be insertion order - which I guess you might be able to argue is randomised in a way...

make_prg/utils/seq_utils.py
```diff
@@ -234,3 +237,46 @@ def get_consensus_from_MSA(alignment: MSA) -> str:
             consensus_string_as_list.append(column.pop())
     consensus_string = "".join(consensus_string_as_list)
     return consensus_string
+
+
+def get_majority_consensus_from_MSA(alignment: MSA) -> str:
+    """
+    Produces a consensus string just based on the major base for each column.
+    Note: alignment needs to be upper-cased bases
```
mbhall88, line 245: Why?

make_prg/utils/seq_utils.py
leoisl: Yes, you are right, it is better to randomly select. Done in https://github.com/iqbal-lab-org/make_prg/pull/60/commits/2089324c0e2bfa68ae21a002e449b81642c560e9

make_prg/utils/seq_utils.py
leoisl: This was not a good choice, fixed in https://github.com/iqbal-lab-org/make_prg/pull/60/commits/2089324c0e2bfa68ae21a002e449b81642c560e9

mbhall88 (approved):

iqbal-lab (approved):

Issue #58, opened 2023-06-07
Ignore `N`s that might have slipped through when updating PRGs


leoisl:
When running the 4-way pipeline, updating the E coli PRG with illumina data, I got this error:
```
Traceback (most recent call last):
  File "/usr/local/bin/make_prg", line 10, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/make_prg/__main__.py", line 94, in main
    args.func(args)
  File "/usr/local/lib/python3.9/site-packages/make_prg/subcommands/update.py", line 182, in run
    denovo_variants_db = DenovoVariantsDB(
  File "/usr/local/lib/python3.9/site-packages/make_prg/update/denovo_variants.py", line 548, in __init__
    locus_name_to_denovo_loci = self._get_locus_name_to_denovo_loci()
  File "/usr/local/lib/python3.9/site-packages/make_prg/update/denovo_variants.py", line 532, in _get_locus_name_to_denovo_loci
    return self._get_locus_name_to_denovo_loci_core(filehandler)
  File "/usr/local/lib/python3.9/site-packages/make_prg/update/denovo_variants.py", line 522, in _get_locus_name_to_denovo_loci_core
    variants = self._read_variants(
  File "/usr/local/lib/python3.9/site-packages/make_prg/update/denovo_variants.py", line 495, in _read_variants
    denovo_variant = cls._read_DenovoVariant(
  File "/usr/local/lib/python3.9/site-packages/make_prg/update/denovo_variants.py", line 477, in _read_DenovoVariant
    denovo_variant = DenovoVariant(
  File "/usr/local/lib/python3.9/site-packages/make_prg/update/denovo_variants.py", line 41, in __init__
    DenovoVariant._param_checking(
  File "/usr/local/lib/python3.9/site-packages/make_prg/update/denovo_variants.py", line 59, in _param_checking
    DenovoVariant._check_sequence_is_composed_of_ACGT_only(alt)
  File "/usr/local/lib/python3.9/site-packages/make_prg/update/denovo_variants.py", line 85, in _check_sequence_is_composed_of_ACGT_only
    raise DenovoError(f"Found a non-ACGT seq ({seq}) in a denovo variant")
make_prg.update.denovo_variants.DenovoError: Found a non-ACGT seq (N) in a denovo variant
```

[...]

PR #61, opened 2023-07-17
Local random object and Skipping malformed denovo variants

leoisl:
This PR fixes two small issues:
1. Fixes a bug introduced in the previous PR, by using a local random object instead of the global random module. In Python, the random module is not process-safe or thread-safe by default. It uses a global instance of the pseudorandom number generator, which is shared between all uses of the random module in the same Python process. It's not designed to handle the concurrent usage scenario well, which is what we have in `make_prg`. Basically, whenever we need to `get_majority_consensus_from_MSA()`, we set the seed of the random object WRT the sequences of the alignment, so that we get the exact same consensus given the same alignment. But we process multiple MSAs at the same time, which could cause issue, e.g. setting `random.seed()` of one MSA while generating the consensus of another MSA (I was having issues with integration tests not being reproducible). We now use a local random object that prevents this issue;
2. We now skip denovo variants that for some very rare reason might have `N` in it (closes https://github.com/iqbal-lab-org/make_prg/issues/58);

Most of the changed files in this PR is updating the zip files in integration tests


leoisl:
CI is failing trying to install scikit-learn for some obscure reason I could not fix yet...

Tests run and pass on my end, total coverage is 99.74%, so it has not decreased:
```
Required test coverage of 98% reached. Total coverage: 99.74%
=========================================================================================== 494 passed, 1 skipped, 4 warnings in 58.95s ===========================================================================================
```

iqbal-lab (approved):
Approved!

mbhall88 (approved):

